PR #8, opened 2020-10-02
do not install dashboards if they are disabled

jmazzitelli:
If the user disabled dashboards, the helm chart will no longer install them. Note that helm will always install the CRD - the crds directory is not templated. So it is always installed.

see https://github.com/kiali/kiali/issues/3285

related operator PR: https://github.com/kiali/kiali-operator/pull/154


jotak (approved):
LGTM

PR #12, opened 2020-10-23
Fix wildcard usage when deleting files

FilipB:
Using * inside quotes does not expand to a list of files so older files
from previous runs were not deleted.


jmazzitelli (approved):
good catch.

PR #15, opened 2020-10-30
Support logger configuration (format, level)

MarianZoll:


kiali-server/values.yaml
```diff
@@ -39,6 +39,9 @@ deployment:
   image_pull_secrets: []
   image_version: ${HELM_IMAGE_TAG}
   ingress_enabled: true
+  logger:
+    log_level: info
+    log_format: text
```
jmazzitelli, line 44: trivial change - but all peer settings are sorted alphabetically here. So change the order. Also wrap string values in quotes, again for consistency.

```suggestion
    log_format: "text"
    log_level: "info"
```

kiali-server/templates/_helpers.tpl
```diff
@@ -32,6 +32,17 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Identifies the log_level with the old verbose_mode and the new log_level considered.
+*/}}
+{{- define "kiali-server.logLevel" -}}
+{{- if .Values.deployment.verbose_mode -}}
+{{- tpl .Values.deployment.verbose_mode . -}}
```
jmazzitelli, line 40: I'm just curious why you used "tpl" here but not for using the log_level value below in the else clause.

kiali-server/templates/deployment.yaml
```diff
@@ -91,6 +89,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: LOG_LEVEL
+          value: {{ include "kiali-server.logLevel" . }}
```
jmazzitelli, line 93: should this value be wrapped in quotes?
```suggestion
          value: "{{ include "kiali-server.logLevel" . }}"
```

kiali-server/templates/deployment.yaml
```diff
@@ -91,6 +89,10 @@ spec:
...
+        - name: LOG_FORMAT
+          value: {{ .Values.deployment.logger.log_format }}
```
jmazzitelli, line 95: I would also prefer this wrapped in quotes unless there is a reason why it is not. I think we know log_format will always be a string.
```suggestion
          value: "{{ .Values.deployment.logger.log_format }}"
```

jmazzitelli (dismissed):
some trivial things, and a question.

kiali-server/templates/_helpers.tpl
MarianZoll: I copied the logic from one of my other projects where the first value could be a Helm argument as well. Thats why `tpl` would render the `verbose_mode` like any other Helm argument in a template in the context of the second argument `.`. Actually we do not need this in this use-case, I am removing the `tpl` function.

jmazzitelli (approved):
This is good. I will put an approval on this PR, but will add a "do not merge" label so we do not merge this until all the other PRs are ready (see https://github.com/kiali/kiali/pull/3382 and https://github.com/kiali/kiali-operator/pull/187)

PR #20, opened 2020-11-25
explicitly define ANSIBLE_CONFIG so an admin can more easily change it

jmazzitelli:
this goes with https://github.com/kiali/kiali/issues/3469

PR #22, opened 2020-12-01
Be able to turn on profiler via helm setting.

jmazzitelli:
This is going to be useful when running molecule tests and we want to see the profiling data. Just pass in `--set debug.enableProfiler=true` when installing via helm.

goes with https://github.com/kiali/kiali/issues/3469 and https://github.com/kiali/helm-charts/pull/20

part of fix for: https://github.com/kiali/kiali/issues/3489


jshaughn (approved):
LOK

PR #28, opened 2020-12-22
Release v1.28.1

israel-hdez:


jmazzitelli (changes requested):
Before we merge this - can you add (either in the git commit or as a comment in this PR) what git commit these tar balls came from? We should have this just to be able to track what code the tarballs came from. Did this come from master? or from 1.28.0 tag? I don't see a commit where VERSION in Makefile was set to v1.28.1 so it will be confusing in the future to know where this build came from.

israel-hdez:
@jmazzitelli It's added in the commit message

jmazzitelli (approved):

PR #29, opened 2021-01-08
add tokenreview to support login check for headers

mlbiam:
additional capability for https://github.com/kiali/kiali/issues/3207

### DIFF
--- a/kiali-server/templates/role.yaml
+++ b/kiali-server/templates/role.yaml
@@ -99,8 +99,7 @@ rules:
   - list
   - patch
   - watch
-- apiGroups:
-  - authentication.k8s.io
+- apiGroups: ["authentication.k8s.io"]
   resources:
   - tokenreviews
   verbs:
--- a/kiali-server/templates/role.yaml
+++ b/kiali-server/templates/role.yaml
@@ -99,4 +99,10 @@ rules:
   - list
   - patch
   - watch
+- apiGroups:
+  - authentication.k8s.io
+  resources:
+  - tokenreviews
+  verbs:
+  - create
 ...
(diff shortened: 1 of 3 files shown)


kiali-server/templates/role.yaml
```diff
@@ -99,4 +99,10 @@ rules:
   - list
   - patch
   - watch
+- apiGroups:
+  - authentication.k8s.io
```
jmazzitelli, line 103: Maintain consistency with the rest of this file - specify apiGroups using `[ ]` notation.
```suggestion
- apiGroups: ["authentication.k8s.io"]
```

jmazzitelli (changes requested):
minor change request

kiali-server/templates/role.yaml
mlbiam: done

jmazzitelli (changes requested):
Because the server now requires a new permission, the operator has to be given this permission also (because the operator will be creating Roles for Kiali and in order to give a permission one has to already have that permission first).

So, this permission needs to also be added to the operator helm chart as well.

So you will need to add this to the bottom here: https://github.com/kiali/helm-charts/blob/master/kiali-operator/templates/clusterrole.yaml#L297

jmazzitelli (changes requested):
Also missing the permission in the role_viewer template.

Read the DEVELOPING.adoc section on "[Are you altering a kiali server role permission](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission)" for details on all the things you need to do when adding a permission.

jmazzitelli (dismissed):
As per [checklist](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission), you are still missing this:
* Update the [Server Helm Chart role-viewer.yaml](https://github.com/kiali/helm-charts/tree/master/kiali-server/templates/role-viewer.yaml)

jmazzitelli (approved):
LGTM

PR #33, opened 2021-01-26
Fix bad replacement when releasing

israel-hdez:
When releasing a `minor` version, a replacement is done to Makefiles to set the
version being released. This replacement is missing the "v" character
used in container image tags, leading to a reference to the wrong image
tags in the resulting helm charts.

This is adding the missing "v".

Fixes kiali/kiali#3617


jmazzitelli (approved):
lgtm

PR #34, opened 2021-02-03
the name of the role being bound to might be different if the user specified fullnameOverride

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3667


jshaughn (approved):
LOK

PR #35, opened 2021-02-03
switch to CRD v1

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/1757

### DIFF
--- a/kiali-operator/crds/crds.yaml
+++ b/kiali-operator/crds/crds.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This is ready for review and testing. Quick testing on my end shows this working (using the helm charts, but the CRDs in the helm charts are the same ones this PR is putting in the operator bundles).

jmazzitelli:
all molecule tests pass with these new CRDs

jmazzitelli:
Note that helm3 does not provide any support for upgrading CRDs - it is up to the admin user to uninstall and then re-install using the new helm chart to pick up the new crd. See: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

jshaughn (approved):
LOK, needed as v1beta1 deprecation end has been announced for Kube 1.22.

PR #39, opened 2021-02-15
Restore CI

israel-hdez:
Since TravisCI is no longer enabled, use GitHub actions for CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: Helm charts
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    branches:
+      - master
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,19 @@
+name: Snyk scan
```
jmazzitelli, line 1: Is this related to "Snyk"?

israel-hdez: Ops! Nope.
Let me change it.

Also, I think there is a syntax error, because I see nothing is running.

israel-hdez:
I'm not sure why it's not running here.
In my fork it runs: https://github.com/israel-hdez/helm-charts/pull/3

israel-hdez:
perhaps since it's new, this needs to be merged?

israel-hdez:
I think this needs to be merged before it can work. I do see a blank check in the "Checks" tab:

image

jmazzitelli (approved):

PR #41, opened 2021-02-15
Fix the Ingress apiVersion so it isn't a deprecated version.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3706

To test, first build with `make build-helm-charts` and then run `helm template -n istio-system kiali-server _output/charts/kiali-server-1.36.0-SNAPSHOT.tgz` and look at the generated Ingress and see that it is using the appropriate api_version for your cluster (it will probably use the v1). You could also install it via `helm install -n istio-system kiali-server _output/charts/kiali-server-1.36.0-SNAPSHOT.tgz`

### DIFF
--- a/kiali-server/templates/ingress.yaml
+++ b/kiali-server/templates/ingress.yaml
@@ -1,7 +1,11 @@
 {{- if not (.Capabilities.APIVersions.Has "route.openshift.io/v1") }}
 {{- if .Values.deployment.ingress_enabled }}
 ---
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+apiVersion: networking.k8s.io/v1
+{{- else }}
 apiVersion: networking.k8s.io/v1beta1
+{{- end }}
 kind: Ingress
 metadata:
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Hmm... this v1 version does not look to be available in k8s versions before 1.19. Before merging, need to figure out if this is a problem. I suspect it is - I think k8s versions 1.18 and lower might still be used and not EOL'ed.

jmazzitelli:
Moved to draft - only merge this when we know we only support k8s environments at the 1.19 version and up.

jmazzitelli:
coded this up so it will support clusters that have the v1 OR v1beta (if there is v1 available, it will use it; otherwise it will fallback and use the v1beta1 deprecated version).

jmazzitelli:
@nrfox added you to review this since you already did the other once. See the description for test procedures - its simple :)

nrfox (approved):
:+1: tested on both kube 1.18 and 1.21

pbajjuri20 (approved):
I run the regression test, everything works perfectly on OpenShift. LGTM

PR #45, opened 2021-02-24
Helm v3 expects a default value for all keys

Lerring:
On Helm v3 I'm unable to use the chart with metrics_enabled="true" . 

`Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in Deployment.spec.template.metadata.annotations.prometheus.io/port`

### DIFF
--- a/kiali-server/templates/deployment.yaml
+++ b/kiali-server/templates/deployment.yaml
@@ -30,7 +30,7 @@ spec:
         prometheus.io/port: {{ .Values.server.metrics_port | quote }}
         {{- else }}
         prometheus.io/scrape: "false"
-        prometheus.io/port: null
+        prometheus.io/port: ""
         {{- end }}
         kiali.io/runtimes: go,kiali
         {{- if .Values.deployment.pod_annotations }}


jmazzitelli:
@Lerring  Are you sure you mean metrics_enabled of "true"? No one has ever seen this error before and [ "metrics_enabled" of "true" is the default](https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L78)- so everyone that has been using this helm chart has not seen this issue (indeed, with that setting equal to true (again, the default) the "null" annotation isn't even processed - so this line that is changed in the PR has no apparent effect if metrics_enabled == true)

Perhaps you meant you set it to "false" and you see an error?

Lerring:
Hi @jmazzitelli apologies, I got them mixed up. This is upon setting to false. Like so:

```
helm install -f values.yaml  \
  --namespace istio-system \
  --set auth.strategy="anonymous" \
  --set server.metrics_enabled="false" \
  --set deployment.ingress_enabled="false" \
  --repo https://kiali.org/helm-charts \
  kiali-server \
  kiali-server
```

jmazzitelli:
I can confirm an error occurs if you disable metrics:

```
$ helm install --namespace istio-system --set server.metrics_enabled=false --repo https://kiali.org/helm-charts kiali-server kiali-server

Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in Deployment.spec.template.metadata.annotations.prometheus.io/port
```

Lerring:
> I can confirm an error occurs if you disable metrics:


Thanks for confirming this

jmazzitelli (approved):
This is good. With "scrape" set to "false", Prometheus should not care about an empty string of port - and that port will be non-empty if you apply the helm chart later with metrics_enabled=true. So this should be fine.

PR #55, opened 2021-03-25
Copy the console files to a memory-backed volume, which allows Kiali to be run read-only-root-dir

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3810


jmazzitelli:
Some quick testing shows this to be working. I will put on the "do not merge" label because I prefer this to not get into tomorrow's release. I want this change to soak in the dev community for a couple weeks before we put it in the release after that.

jmazzitelli:
closing this PR - we can always resurrect it if we want to come back to this implementation, but it seems like we want to try something else.

PR #61, opened 2021-04-12
Be able to disable ad hoc images.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3865

### DIFF
--- a/kiali-operator/templates/deployment.yaml
+++ b/kiali-operator/templates/deployment.yaml
@@ -60,6 +60,8 @@ spec:
               fieldPath: metadata.namespace
         - name: ALLOW_AD_HOC_KIALI_NAMESPACE
           value: {{ .Values.allowAdHocKialiNamespace | quote }}
+        - name: ALLOW_AD_HOC_KIALI_IMAGE
+          value: {{ .Values.allowAdHocKialiImage | quote }}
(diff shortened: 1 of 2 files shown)


jmazzitelli:
@israel-hdez please review - this addresses the image override stuff you mentioned recently.

israel-hdez (approved):
This part looks fine.

Although I see the default value for `allowAdHocKialiNamespace` (the one already existent) is `true`. While the default value of the OperatorHub metadata is set to `false` (see: https://github.com/kiali/kiali-operator/blob/a8dd1aea9a9cee5d5b575386cb4cce4d9adac4ad/manifests/kiali-community/1.33.0/kiali.v1.33.0.clusterserviceversion.yaml#L274-L275).

I just wanted to mention that, as I see different behaviors depending on the installation channel.

jmazzitelli:
> I just wanted to mention that, as I see different behaviors depending on the installation channel.

Yes. This is on purpose, as I believe most people that use the operator helm chart are using it in a manner that requires this setting to be true. We are now giving people the option to turn it off if they want, but for backward compatibility with the way people are using this today, we enable it by default. (I believe people are putting the Kiali CR in kiali-operator namespace but want to deploy Kiali in istio-system.)

PR #67, opened 2021-04-23
be able to customize the length of the ansible profile report

jmazzitelli:
I was using this as part of debugging https://github.com/kiali/kiali/issues/3892 - but this can be merged independent of that.

When you enable the profiler, show the top 100 tasks, not just the default 20


jshaughn (approved):
LOK as long as you really meant to have TASKS_TASK

jmazzitelli:
> LOK as long as you really meant to have TASKS_TASK

This isn't my env var - this is a env var declared by and used by the ansible operator SDK. See ansible docs here: https://docs.ansible.com/ansible/latest/collections/ansible/posix/profile_tasks_callback.html#parameters

PR #71, opened 2021-05-04
Add istio options needed for canary upgrades to values

nrfox:
Adds istio options needed for canary upgrades to kiali-server helm chart values.

Part of https://github.com/kiali/kiali/issues/2884.

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -65,6 +65,12 @@ deployment:
 external_services:
   custom_dashboards:
     enabled: true
+  istio:
+    # These three values control which istio controlplane the kiali server talks to.
+    # Useful for canary Istio upgrades.
+    config_map_name: "istio"
+    istio_sidecar_injector_config_map_name: "istio-sidecar-injector"
+    istiod_deployment_name: "istiod"
 
 identity: {}
   #cert_file:


kiali-server/values.yaml
```diff
@@ -65,6 +65,12 @@ deployment:
 external_services:
   custom_dashboards:
     enabled: true
+  istio:
+    # These three values control which istio controlplane the kiali server talks to.
+    # Useful for canary Istio upgrades.
+    config_map_name: "istio"
+    istio_sidecar_injector_config_map_name: "istio-sidecar-injector"
+    istiod_deployment_name: "istiod"
```
jmazzitelli, line 73: Do we need to define these here? Note on line 14 we indicate that we only define settings here in values.yaml that are actually used in the helm charts themselves (otherwise, we'd be forced to make a virtual copy of our kiali_cr.yaml here which kinda makes it a maintenance headache to now have to maintain kiali_cr.yaml and this yaml together in lock step).

If we don't use these external_services.istio section of settings anywhere in our server helm chart, then we should not do this change.

nrfox: How would a user update settings for their kiali server (data in the configmap) through the helm chart otherwise?

jmazzitelli: > How would a user update settings for their kiali server (data in the configmap) through the helm chart otherwise?

Just set it in the helm command via `--set` and it "just works" (e.g. `--set external_services.istio.config_map_name=foo`, rinse-repeat for the other 2).

In our documentation we would mention these settings but we do not need them explicitly declared here.

This is where the [magic happens](https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/configmap.yaml#L12)... you can read about [`omit` here](https://helm.sh/docs/chart_template_guide/function_list/#omit)

nrfox: Ah that's great! I'll close this PR and just include that in the docs change for upgrades.

nrfox:
Closing since passing these values through to the configmap is already supported by the helm chart.

PR #73, opened 2021-05-12
Remove quote from apiVersion of Kiali Server

KeisukeYamashita:
## What

I want to remove the `""` from `apiVersion` because other CRDs and Kubernetes native APIs does not have it.

### DIFF
--- a/kiali-server/templates/dashboards/envoy.yaml
+++ b/kiali-server/templates/dashboards/envoy.yaml
@@ -1,6 +1,6 @@
 {{- if (include "kiali-server.isDashboardEnabled" .) }}
 ---
-apiVersion: "monitoring.kiali.io/v1alpha1"
+apiVersion: monitoring.kiali.io/v1alpha1
 kind: MonitoringDashboard
 metadata:
   name: envoy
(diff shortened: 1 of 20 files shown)


jmazzitelli:
Are the original files breaking something? I'm not sure what bug this is fixing or what enhancement this is giving us. The original files contain perfectly valid YAML and Kubernetes allows for quoting string values. I question "change for the sake of change" without some added benefit.

If you want to change Kiali installation components, please refer to [the developer docs here](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#how-to-change-resources-or-configuration) -- specifically, changing the server helm chart isn't enough - there are [copies of these files](https://github.com/kiali/kiali-operator/tree/master/roles/default/kiali-deploy/templates/dashboards) also in the operator for use by the operator installation mechanism.

KeisukeYamashita:
Thanks for your reply. 

I want to clearify that it's not breaking anything but I can find any reasonable reason to quote this value. It's for sure a valid YAML and I sent out a PR to correct the problem because it was uncomfortable and align with all other objects. I have CRDs for Datadog, ArgoCD, etc but Kaili is only one that the value is quoted.

Since it won't break something, even if it is refered by other components, this change won't break anything as far as I know.

jmazzitelli:
FWIW: I'm going to be removing those quotes in another PR that is also touching these monitoring dashboard yamls - so this will be done once this PR is merged --> https://github.com/kiali/kiali-operator/pull/313

KeisukeYamashita:
Great, Thank you very much!!!

PR #75, opened 2021-05-15
[helm] set app.kubernetes.io/instance label appropriately, define the new deployment.instance_name value

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4018

This only touches the kiali-server helm chart.

If you leave all defaults as-is, this works as before and you will not notice any differences. The only difference is if you want to set `fullnameOverride`.

To simulate the way the operator works, the kiali-server helm chart now uses `deployment.instance_name` rather than the now-deprecated `fullnameOverride`. For backwards compatibility, if `fullnameOverride` is not `kiali` but `deployment.instance_name` remains its default value of `kiali`, then `fullnameOverride` is used, otherwise `deployment.instance_name` is used. The old `nameOverride` is deprecated and now ignored.

A second commit in this PR merely removes the quotes from the dashboard apiVersion fields for consistency to all other resources.

### DIFF
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -1,27 +1,17 @@
 {{/* vim: set filetype=mustache: */}}
 
 {{/*
-Expand the name of the chart.
-*/}}
-{{- define "kiali-server.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
-{{- end }}
-
-{{/*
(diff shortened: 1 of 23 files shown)


jmazzitelli:
This is a companion to https://github.com/kiali/kiali-operator/pull/313

I did a quick test and this works.

jmazzitelli:
To test - install via helm by passing in `--set deployment.instance_name` and see that the resources have their names prefixed with that setting's value (not "kiali").

PR #76, opened 2021-05-17
more security to the kiali pod

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4023


israel-hdez (approved):
I see manifest is equivalent to the operator PR.

I tried this with:

`kubectl apply --validate=true --dry-run=client -f <(helm template kiali-test ./kiali-server)`

All good!

PR #82, opened 2021-06-18
Always mount the kiali-cabundle volume/configmap

israel-hdez:
For Kubernetes, it is marked as optional. For OpenShift, it is still
mandatory.

Related operator PR: https://github.com/kiali/kiali-operator/pull/333

Related kiali/kiali#4050


jmazzitelli (approved):
i didn't run this, but this chance looks correct. Now we always define this cabundle stuff, but it is optional on kubernetes since on k8s it isn't always used.

PR #89, opened 2021-07-29
Grant permissions to kiali operator SA to read Secrets

leandroberetta:
Needed by [#4227](https://github.com/kiali/kiali/pull/4227)

### DIFF
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -33,10 +33,20 @@ rules:
   resources:
   - secrets
   verbs:
-  - get
   - create
   - list
   - watch
+{{- if gt (len .Values.secretReader) 0 }}
+- apiGroups: [""]
+  resourceNames:
+  {{- range .Values.secretReader }}
+  - {{ . }}
+  {{- end }}
+  resources:
+  - secrets
+  verbs:
+  - get
+{{- end }}
 - apiGroups: [""]
   resourceNames:
   - kiali-signing-key
@@ -183,7 +193,7 @@ rules:
   - watch
 # The permissions below are for Kiali itself; operator needs these so it can escalate when creating Kiali's roles
 - apiGroups: [""]
-  resources:  
+  resources:
   - configmaps
   - endpoints
   - pods/log
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -182,8 +182,7 @@ rules:
   - watch
 # The permissions below are for Kiali itself; operator needs these so it can escalate when creating Kiali's roles
 - apiGroups: [""]
-  resources:
-  - secrets
+  resources:  
   - configmaps
   - endpoints
   - pods/log
@@ -311,4 +310,14 @@ rules:
   - tokenreviews
   verbs:
   - create
+- apiGroups: [""]
+  resourceNames:
+  - cacerts
+  - istio-ca-secret
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
 ...
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -33,6 +33,7 @@ rules:
   resources:
   - secrets
   verbs:
+  - get
   - create
   - list
   - watch
@@ -310,14 +311,4 @@ rules:
   - tokenreviews
   verbs:
   - create
-- apiGroups: [""]
-  resourceNames:
-  - cacerts
-  - istio-ca-secret
-  resources:
-  - secrets
-  verbs:
-  - get
-  - list
-  - watch
 ...
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -183,6 +183,7 @@ rules:
 # The permissions below are for Kiali itself; operator needs these so it can escalate when creating Kiali's roles
 - apiGroups: [""]
   resources:
+  - secrets
   - configmaps
   - endpoints
   - pods/log
--- a/kiali-operator/values.yaml
+++ b/kiali-operator/values.yaml
@@ -37,9 +37,13 @@ watchNamespace: ""
 # Note that this will be overriden to "true" if cr.create is true and cr.spec.deployment.accessible_namespaces is ['**'].
 clusterRoleCreator: true
 
-# Set to a list of secrets if you want the operator to be able to read those secrets across the cluster. This is necessary
-# if you want to support Kiali CRs with spec.kiali_feature_flags.certificates_information_indicators.enabled = true.
-# If you set this to an empty list, the operator will not be given permission to read secrets outside of the control plane namespace.
+# Set to a list of secrets in the cluster that the operator will be allowed to read. This is necessary if you want to
+# support Kiali CRs with spec.kiali_feature_flags.certificates_information_indicators.enabled=true.
+# The secrets in this list will be the only ones allowed to be specified in any Kiali CR (in the setting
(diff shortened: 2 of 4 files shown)


jmazzitelli (changes requested):
This is a problem. We are going to need a project-wide discussion before we merge. Do not merge until this is vetted by the team and community.

We specifically have been avoiding adding this "read any secret" permission. People do not like it (we recently had a community member complain about our current ability to read secrets in the control plane namepace). This is a very, very bad (potential) security hold that kiali would be opening up.

We need to avoid this permission. If you need to read a specific secret, we can do so by name as we do here: https://github.com/kiali/helm-charts/blob/master/kiali-operator/templates/clusterrole.yaml#L41-L52

kiali-operator/templates/clusterrole.yaml
```diff
@@ -184,7 +185,7 @@ rules:
   - watch
 # The permissions below are for Kiali itself; operator needs these so it can escalate when creating Kiali's roles
 - apiGroups: [""]
-  resources:
+  resources:
```
jmazzitelli, line 188: this is a no-op change (spaces at EOL?) - revert
```suggestion
  resources:
```

kiali-server/templates/role-controlplane.yaml
```diff
@@ -14,4 +14,14 @@ rules:
...
+- apiGroups: [""]
+  resourceNames:
+  - cacerts
+  - istio-ca-secret
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
jmazzitelli, line 26: This section isn't using the new config setting. This instead should be:
```suggestion
{{- if .Values.kiali_feature_flags.certificates_information_indicators.enabled }}
- apiGroups: [""]
  resourceNames: {{ .Values.kiali_feature_flags.certificates_information_indicators.secrets }}
  resources:
  - secrets
  verbs:
  - get
  - list
  - watch
{{- end }}
```
I'm not exactly sure of the syntax to get that list value set in `resourceNames` but you get the idea.

jmazzitelli (changes requested):

kiali-operator/templates/clusterrole.yaml
```diff
@@ -35,6 +35,7 @@ rules:
   resources:
   - secrets
   verbs:
+  - get
```
jmazzitelli, line 38: see my comment coming about `secretReader: true` setting

jmazzitelli:
I think we want to add another flag to let people turn off the global "get" permission if they don't want to even grant the operator that permission.

So we want an analogous setting as this: https://github.com/kiali/helm-charts/blob/v1.38.0/kiali-operator/values.yaml#L35-L38
which is used here: https://github.com/kiali/helm-charts/blob/v1.38.0/kiali-operator/templates/clusterrole.yaml#L102-L105

For example:

values.yaml:
```
# Set to true if you want the operator to be able to read secrets across the cluster. This is necessary
# if you want to support Kiali CRs with spec.kiali_feature_flags.certificates_information_indicators.enabled = true.
# If you set this to false, the operator will not be given permission to read secrets outside of the control plane namespace.
secretReader: true
```

templates/clusterrole.yaml:
```
  {{- if .Values.secretReader }}
  - get
  {{- end }}
```

This should help users that are paranoid and don't want to even the operator to be granted permissions to read all secrets. I don't think the operator has global/cluster-wide ability to read secrets today - this keeps it that way if the user wants.

jmazzitelli:
I created PR https://github.com/leandroberetta/helm-charts/pull/1 that adds the stuff I talk about above. This makes me feel alittle better :)

kiali-server/values.yaml
```diff
@@ -73,6 +73,11 @@ identity: {}
   #private_key_file:
 
 kiali_feature_flags:
+  certificates_information_indicators:
+   enabled: true
+   secrets:
+   - istio-ca-secret
+   - cacerts
```
jmazzitelli, line 80: can you swap this to sort them alphabetically?
It's stupid, I know - but when we do diff's on these lists, its nice when things are sorted because the diffs won't just show different orders of things.
so this is trivial:
```suggestion
   - cacerts
   - istio-ca-secret
```

jmazzitelli (changes requested):
trivial change requested

kiali-operator/values.yaml
```diff
@@ -37,6 +37,11 @@ watchNamespace: ""
 # Note that this will be overriden to "true" if cr.create is true and cr.spec.deployment.accessible_namespaces is ['**'].
 clusterRoleCreator: true
 
+# Set to a list of secrets if you want the operator to be able to read those secrets across the cluster. This is necessary
+# if you want to support Kiali CRs with spec.kiali_feature_flags.certificates_information_indicators.enabled = true.
+# If you set this to an empty list, the operator will not be given permission to read secrets outside of the control plane namespace.
```
jmazzitelli, line 42: I would like to reword this now that I'm looking at it and reading what I wrote. That last sentence isn't accurate because the operator will still have permission to read a secret named "kiali-signing-key" found anywhere. And it isn't true that the operator can't read secrets "outside of the control plane namespace' because the operator doesn't know ahead of time what control  plane namespaces there are or will be. So let's reword this to make it more clear and specific as to what this setting does:

```suggestion
# Set to a list of secrets in the cluster that the operator will be allowed to read. This is necessary if you want to
# support Kiali CRs with spec.kiali_feature_flags.certificates_information_indicators.enabled=true.
# The secrets in this list will be the only ones allowed to be specified in any Kiali CR (in the setting
# spec.kiali_feature_flags.certificates_information_indicators.secrets).
# If you set this to an empty list, the operator will not be given permission to read any additional secrets
# found in the cluster, and thus will only support a value of "false" in the Kiali CR setting
# spec.kiali_feature_flags.certificates_information_indicators.enabled.
```

jmazzitelli (changes requested):
We need to update the comment for secretReader to make it more clear what that is.

kiali-server/templates/role-controlplane.yaml
```diff
@@ -14,4 +14,17 @@ rules:
...
+  - {{ . }}
+  {{- end }}
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
 ...
+{{- end }}
```
jmazzitelli, line 30: This is in the wrong spot. If the feature flags.enabled is false, then the end of file "..." marker isn't included. We want that included all the time, whether the enabled flag is true or not.

So make the final two lines of this file:

```
{{- end }}
...
```

jmazzitelli (dismissed):
sorry - just noticed one more thing wrong.

kiali-server/templates/role-controlplane.yaml
leandroberetta: Right, thanks, fixed.

jmazzitelli:
I just tested by setting `--set "secretReader={cacerts}"` - and things worked as expected. The default Kiali CR settings cause a failure because kiali can't be granted access to read istio-ca-secret. However, when I then decided to not give the operator any get-secret perms, helm spit out an error and the operator did not get installed:

```
$ helm install --set "secretReader={}" --set "image.repo=localhost:5000/kiali/kiali-operator" --set image.tag=dev --set allowAdHocKialiImage=true -n kiali-operator kiali-operator _output/charts/kiali-operator-1.39.0-SNAPSHOT.tgz 

Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(ClusterRole.rules[3].resourceNames): unknown object type "nil" in ClusterRole.rules[3].resourceNames[0]
```

I am investigating why that happened.

jmazzitelli (dismissed):
after a bunch of testing, this all looks good.

jmazzitelli (approved):
re-approve

PR #97, opened 2021-08-10
Add feature flag for clustering/federation

israel-hdez:
For now, this flag is only allowing or preventing creation of a role and
a rolebinding that grants privileges for reading some secrets in the
Istio namespace. There is no effect on Kiali server side, because it is
failing gracefully. In case it's needed later, this flag is being passed
to Kiali in its ConfigMap.

Default value is `true` for backwards compatibility.

Related kiali/kiali#4147

### DIFF
--- a/kiali-server/templates/role-controlplane.yaml
+++ b/kiali-server/templates/role-controlplane.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.kiali_feature_flags.clustering }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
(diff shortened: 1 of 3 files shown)


israel-hdez:
@jmazzitelli updated to work similarly as the operator in https://github.com/kiali/kiali-operator/pull/395

jmazzitelli (approved):
this is good. when passing in `--set kiali_feature_flags.clustering.enabled=false` I get this:

```
$ kubectl get role -n istio-system kiali-controlplane 
NAME                 CREATED AT
kiali-controlplane   2021-08-12T20:04:09Z

$ kubectl get role -n istio-system kiali-controlplane -o jsonpath={.rules}
<nil>
```

PR #98, opened 2021-08-19
removing events and pvc

prary:
fixes: https://github.com/kiali/kiali/issues/4290 (UPDATED by mazz with the correct url)

removing events and pvc

### DIFF
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -10,8 +10,6 @@ rules:
   resources:
   - configmaps
   - endpoints
-  - events
-  - persistentvolumeclaims
   - pods
   - serviceaccounts
   - services


jshaughn:
Thanks @prary for the PR.  Can you add to your desription the tests that you've performed and also did you read through https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission-that-applies-to-resources-in-all-accessible-namespaces to ensure you've hit all of the touch points?

cc @jmazzitelli

jmazzitelli:
> did you read through https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission-that-applies-to-resources-in-all-accessible-namespaces to ensure you've hit all of the touch points?

I did mention that here: https://github.com/kiali/kiali/issues/4290#issuecomment-901918572

jmazzitelli:
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-ossm/manifests/kiali.clusterserviceversion.yaml#L292-L293
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-community/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L251-L252
* https://github.com/kiali/kiali-operator/blob/master/manifests/kiali-upstream/1.39.0/manifests/kiali.v1.39.0.clusterserviceversion.yaml#L251-L252

those last two are probably going to need to change in two days since this PR and related changes won't make it in the 1.39 release due to go out tomorrow. So when we release 1.39 we'll release those CSV files as-is and then we'll create a 1.40 version of those files - those would be the ones that need to change.

But I think that's the only other 3 places for this change since this only will affect the operator role.

jmazzitelli:
I will test this PR now with all the molecule tests and see how it works on minikube.

jmazzitelli:
all molecule tests pass with this PR. So these permissions look obsolete and no longer needed.

jmazzitelli:
Too late to get this in for our release tomorrow. We'll merge this next week for inclusion in 1.40 release in 3 weeks. We'll also need to create the 1.40 CSVs for the OLM/operator metadata and then we'll update those as well.

jmazzitelli:
this removes the perms from the OLM metadata: https://github.com/kiali/kiali-operator/pull/403

jmazzitelli (approved):

PR #102, opened 2021-09-06
Ability to use a custom configmap

nustiueudinastea:
Hi folks,

This PR adds the option to use a custom configmap to configure kiali. This ability is useful in certain cases and users won't notice anything when using the default behaviour.

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -26,6 +26,9 @@ auth:
   strategy: ""
 
 deployment:
+  # This is a user defined configmap that should hold the configuration for Kiali. It will be used instead of the default one if set.
+  # The custom configmap should have a data entry for "config.yaml", which will be used by Kiali at startup.
+  configuration_configmap: ""
   # This only limits what Kiali will attempt to see, but Kiali Service Account has permissions to see everything.
   # For more control over what the Kial Service Account can see, use the Kiali Operator
   accessible_namespaces:
--- a/kiali-server/templates/configmap.yaml
+++ b/kiali-server/templates/configmap.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.deployment.configuration_configmap }}
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -23,3 +24,4 @@ data:
     {{- $_ := set $cm.server "web_root" (include "kiali-server.server.web_root" .) }}
     {{- toYaml $cm | nindent 4 }}
 ...
+{{- end}}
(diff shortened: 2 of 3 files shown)


jmazzitelli:
@nustiueudinastea please create an enhancement request issue in the kiali/kiali project via [this link](https://github.com/kiali/kiali/issues/new?assignees=&labels=enhancement&template=feature_request.md&title=) detailing the enhancement request you are proposing (fill in the issue template) and link this PR with that issue in the PR description via "fixes: \<the PR URL>" line.

nustiueudinastea:
I created an issue as requested: https://github.com/kiali/kiali/issues/4334

kiali-server/values.yaml
```diff
@@ -26,6 +26,9 @@ auth:
   strategy: ""
 
 deployment:
+  # This is a user defined configmap that should hold the configuration for Kiali. It will be used instead of the default one if set.
+  # The custom configmap should have a data entry for "config.yaml", which will be used by Kiali at startup.
+  configuration_configmap: ""
```
jmazzitelli, line 31: This should not go under deployment. This is an external configuration setting (specific to the server helm chart itself). These settings under "deployment" mimic those that are in the Kiali CR. This should be a top-level value like `fullnameOverride`.

jmazzitelli (changes requested):

jmazzitelli (changes requested):
De-coupling the configmap like what this PR proposes is going to cause issues - for example, all the defaults that are dynamically determined and stored in the Kiali config map. So this PR is going to have to be changed. One of the important things (because it involves security and because someone submitted a security bulletin/CVE against Kiali before for this) is the signing key default. We have to ensure the signing key is defaulted as a random value if the user doesn't provide one (and the user rarely will provide one). If we do not, we will get flagged with a security violation by the community. There are other dynamically generated non-security values also (some depends on the cluster kind like OpenShift versus non-OpenShift).

To fix this, I think if the helm chart can somehow slurp in the external config map and read in its values as the first thing it does (say, in _helpers template) and use those external configmap settings as helm values, too (just as if you gave them to helm as a values.yaml - i.e. put them in .Values) then things would just work normally (the helm chart would create its own configmap with its own set of dynamically generated values, but with the defaults defined in this external config map).

kiali-server/values.yaml
jmazzitelli: This request-for-change still stands and should be done in conjunction with my other suggested change in the new review-for-change I just submitted (i.e. move this to the top level and then have the helm chart read in this external configmap and store the values in .Values so the helm charts can use the values in the external configmap as its defaults).

israel-hdez:
No activity for 6 months.
Closing this and it's linked issue as _stale_.

PR #84, opened 2021-06-22
No need for pods/proxy subresource so far

xeviknal:
needs https://github.com/kiali/kiali/pull/4108

There is no need to communicate with Istiod pods using the pod/proxy. Kubernetes assigns a PodIP accessible from all nodes of a cluster.

PR #107, opened 2021-09-16
fix kiali roles

zackzhangkai:
When install kiali server alone, instead of operator, it will occurs Warning log such as below:
```
2021-09-16T10:07:04Z WRN GetPodProxyStatus is failing for [namespace: csminstalltest] [pod: reviews-v1-56b6cb7f4c-rq9nf]: unable to proxy Istiod pods. Make sure your Kubernetes API server has access to the Istio control plane through 8080 port
```

Refer to https://github.com/kiali/kiali/issues/3494#issuecomment-771612162, I found the RBAC role is missing pods/proxy permission.

Signed-off-by: zackzhangkai <zhang_kai@cestc.cn>

### DIFF
--- a/kiali-server/templates/role.yaml
+++ b/kiali-server/templates/role.yaml
@@ -11,6 +11,7 @@ rules:
   - configmaps
   - endpoints
   - pods/log
+  - pods/proxy
   verbs:
   - get
   - list


lucasponce:
In the current versions Kiali moved to use "pods/forward" to query the Envoy information to calculate status and other admin APIs.
So, not sure if bringing back that role would be correct or at least I guess it may cause some side effects.

I wonder if you could describe the combination of versions of Kiali you are using and the mechanism to install it, I suspect there should be some mix there.

Thanks for opening the PR and bringing this.

jmazzitelli (changes requested):
No, this isn't right. Please post your full installation command you are using (including all the values.yaml and --set commands). I suspect you are using the current helm charts with an old version of the image.

jmazzitelli:
Here is the PR where we explicitly removed these permissions back in June - https://github.com/kiali/helm-charts/pull/84

I'm closing this PR as it introduces a permission we no longer need and we already explicitly removed.

PR #109, opened 2021-10-01
added hostaliases to kiali-server

ethernoy:
Signed-off-by: Ethern Su <ehaprime@gmail.com>


jmazzitelli (changes requested):
Just looking at this briefly, it seems like it is OK (I need to research that hostAliases setting - that is a new one to me). However, please be advised Kiali users are more than just helm server users. You may need to update the helm chart for the operator and most definitely you need to create a  PR to have this same capability for those that deploy Kiali via the operator and not the server helm chart.

Please refer to these developer checklists to know what things need to be done:
* https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-resource
* https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting

ethernoy:
> Just looking at this briefly, it seems like it is OK (I need to research that hostAliases setting - that is a new one to me). However, please be advised Kiali users are more than just helm server users. You may need to update the helm chart for the operator and most definitely you need to create a PR to have this same capability for those that deploy Kiali via the operator and not the server helm chart.
> 
> Please refer to these developer checklists to know what things need to be done:
> 
> * https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-resource
> * https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting

I can do a PR for operator too, but it will be in the operator repository and not this one. Do you think I should work on that first before continuing on this one?

jmazzitelli:
> I can do a PR for operator too, but it will be in the operator repository and not this one. Do you think I should work on that first before continuing on this one?

Yes, correct - there will be a PR in the operator repo as well as this one. Thanks.

jmazzitelli:
companion operator PR: https://github.com/kiali/kiali-operator/pull/432

kiali-server/templates/deployment.yaml
```diff
@@ -47,6 +47,10 @@ spec:
       - name: {{ . }}
       {{- end }}
       {{- end }}
+      {{- if .Values.deployment.host_aliases }}
+      hostAliases:
+      {{ toYaml .Values.deployment.host_aliases | nindent 8 }}
```
jmazzitelli, line 52: missing the `-` prefix and the indent it 2 too many. Because of this, an extraneous newline appears and the indentation is off:

```
      hostAliases:

        - hostnames:
          ...and the rest...
```

With the suggested change, that extraneous newline is removed and indentation is corrected:

```
      hostAliases:
      - hostnames:
        ...and the rest...
```

```suggestion
      {{- toYaml .Values.deployment.host_aliases | nindent 6 }}
```

With that fixed, I can verify the results:

```
kubectl exec -n istio-system $(kubectl get pod -n istio-system -l app=kiali -o name) -- cat /etc/hosts
```

Will show the host entries in /etc/hosts of the kiali server container:

```
# Kubernetes-managed hosts file.
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback

...

# Entries added by HostAliases.
192.168.1.111	foo.local	bar.local
192.168.1.222	foo2.local	bar2.local
```

jmazzitelli (changes requested):
small changes needed to the helm `toYaml` line - once that change is made, this should all work fine. See my change request for the test procedure (it is just `cat /etc/hosts` in the container to see the host aliases are created).

kiali-server/templates/deployment.yaml
ethernoy: thx for your fix!

jmazzitelli:
Here are the test procedures to confirm this works (I ran these and all works):

1. Pull this PR and build the helm-charts via make:
```bash
make clean build-helm-charts
```
2. Tell helm to output the template yaml, setting the new host_aliases fields:
```bash
helm template -n istio-system kiali-server _output/charts/kiali-server-*-SNAPSHOT.tgz --set deployment.host_aliases[0].ip=192.168.1.123 --set deployment.host_aliases[0].hostnames="{foo.local,bar.local}"
```
3. Confirm the hostAliases field is correctly set in the Deployment. Based on the command in step 2, you should see this in the Deployment yaml:
```yaml
spec:
  template:
    spec:
      hostAliases:
      - hostnames:
        - foo.local
        - bar.local
        ip: 192.168.1.123
```
4. If you want to test further, you can actually deploy this in a cluster that has Istio installed:
```bash
helm install -n istio-system kiali-server _output/charts/kiali-server-*-SNAPSHOT.tgz --set deployment.host_aliases[0].ip=192.168.1.123 --set deployment.host_aliases[0].hostnames="{foo.local,bar.local}"
```
5. Once the Kiali pod is running, output its `/etc/hosts` file:
```bash
kubectl exec -n istio-system $(kubectl get pod -n istio-system -l app=kiali -o name) -- cat /etc/hosts
```
6. Now check that `/etc/hosts` file and confirm it has the custom alias installed. You should see something like this:
```
# Kubernetes-managed hosts file.
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
[...and some other settings...]

# Entries added by HostAliases.
192.168.1.123	foo.local	bar.local
```

jmazzitelli (approved):
tests pass. This looks good.

jmazzitelli:
assigning to @mattmahoneyrh in case QE wants to verify. Test procedures found in previous comment.

PR #111, opened 2021-10-08
Add annotation in case of changing ConfigMap

sly-roar:
If a ConfigMap changes that does not trigger re-deployment of Kiali Server. 
Since the ConfigMap is used for configuring Kiali Server the "rebooting" the Server is required. 
Adding the annotation of kind
`checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}`
helps to force performing rollout of deployment and re-create the pod with Kiali Server.

### DIFF
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -70,7 +70,11 @@ Determine the default web root.
 */}}
 {{- define "kiali-server.server.web_root" -}}
 {{- if .Values.server.web_root  }}
-  {{- .Values.server.web_root | trimSuffix "/" }}
+  {{- if (eq .Values.server.web_root "/") }}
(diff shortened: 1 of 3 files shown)


jmazzitelli:
This has exposed some issues with the way web_root is processed throughout. To fix this PR, we'll need to address those other places too. Just look at : https://github.com/kiali/helm-charts/search?q=web_root&type=code and notice the trimming of the web_root string where it should not be happening (trimming a string `/` results in an empty string, and I don't think that's what we want in most/all of these - that's why the Ingress failed to get created in my test procedures mentioned in my previous comment).

We should go through these and remote the trims where ever we see this `{{ include "kiali-server.server.web_root" . | trimSuffix "/"  }}` -- there is no need to trim here because that included template should have already trimmed the string.

PR #118, opened 2021-11-15
implement some best practices

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/4485

### DIFF
--- a/kiali-operator/values.yaml
+++ b/kiali-operator/values.yaml
@@ -13,7 +13,10 @@ nodeSelector: {}
 podAnnotations: {}
 env: []
 tolerations: []
-resources: {}
+resources:
+  requests:
+    cpu: "10m"
+    memory: "64Mi"
 affinity: {}
 replicaCount: 1
 priorityClassName: ""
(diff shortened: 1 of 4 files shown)


jmazzitelli:
this first commit provides the ability to add custom labels to the Ingress/Route

jmazzitelli:
second commit sets the resource requests. We might also consider setting resource limits, but so far that is not implemented.

jmazzitelli:
third commit sets the resource requests on the server. We might also consider setting resource limits, but so far that is not implemented.

jmazzitelli:
fourth commit adds memory limits to the server pod. We are going to opt not to create memory limits for the operator pod right now

jmazzitelli:
to test - set the helm value `deployment.ingress.additional_labels.foo=bar` and (if on non-OpenShift/minikube/kind) set `spec.deployment.ingress.enabled=true`. When you install, check that the Ingress has the label "foo: bar" and check the server pod's container to ensure it has `resources` defaults:
```yaml
resources:
  requests:
    cpu: 10m
    memory: 64Mi
  limits;
    memory: 1Gi
```
If you test on OpenShift, check that the labels are on the Route (a Route is created on OpenShift, not an Ingress).

Then test by installing the operator via its helm chart and make sure it's pod container also has resources defaults:
```yaml
resources:
  requests:
    cpu: 10m
    memory: 64Mi
```

nrfox (approved):

PR #131, opened 2022-04-06
Add a new CR setting. 

doflamingo721:
Adding a new config which can be used as in-memory replace in env.js file and using `server.web_root` as default if the value of new config is not specified.

**Linked issue:**
https://github.com/kiali/kiali/issues/4459

**Linked PRs:**
kiali: https://github.com/kiali/kiali/pull/4914
kiali-operator: https://github.com/kiali/kiali-operator/pull/510

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -103,6 +103,7 @@ login_token:
 
 server:
   port: 20001
+  redirect_path: ""
   metrics_enabled: true
   metrics_port: 9090
   web_root: ""


jmazzitelli (changes requested):
You do not need this PR. ​We only need to add defaults to the values file if that setting is used elsewhere in the templates. This value is not used anywhere so we do not add it to the values. See: https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L18 where it says `Note that only those values used by the Helm Chart will be here.`

So this PR can be closed.

jmazzitelli:
closing this - this is not needed for the issue at hand. The other two PRs remain: https://github.com/kiali/kiali/pull/4914 and https://github.com/kiali/kiali-operator/pull/510

PR #133, opened 2022-04-26
Release pipeline for Helm charts

leandroberetta:
This is the migrated pipeline for the Helm charts.

Resolves https://github.com/kiali/kiali/issues/4992

Epic: https://github.com/kiali/kiali/issues/4827


jmazzitelli (approved):
I don't see anything wrong but I didn't run any of this. I assume this is tested in your own branch? The one thing that we need to be sure of is the master branch needs to be updated (the `docs/` directory in `master` is where we serve the helm charts from - that's where the http server gets the content). But I see this is done here (see the comment "Switch to `master` branch before updating docs/index.yaml") so that looks good.

leandroberetta:
> I don't see anything wrong but I didn't run any of this. I assume this is tested in your own branch? The one thing that we need to be sure of is the master branch needs to be updated (the `docs/` directory in `master` is where we serve the helm charts from - that's where the http server gets the content). But I see this is done here (see the comment "Switch to `master` branch before updating docs/index.yaml") so that looks good.

@jmazzitelli thanks for the review. I tested this on my fork, the following is an example of the PR created for that:

https://github.com/leandroberetta/helm-charts/pull/6

.github/workflows/release.yml
```diff
@@ -0,0 +1,203 @@
...
+        default: 'auto' 
+        type: choice
+        options:
+          - minor
+          - patch 
+      repository:
+        description: Repository (in owner/repo format)
+        required: true
+        default: kiali/helm-charts
+        type: string
```
israel-hdez, line 18: Like I said in the other PR, I think this input should be removed. It should always be the repo where the workflow was triggered.

.github/workflows/release.yml
```diff
@@ -0,0 +1,203 @@
...
+          RELEASE_TYPE: ${{ github.event.inputs.release_type }}
+          RELEASE_VERSION: ${{ steps.release_version.outputs.release_version }}
+        id: next_version
+        run: |
+          # Drop the version prefix ("v")
+          RELEASE_VERSION=${RELEASE_VERSION:1}
+
+          if [[ $RELEASE_TYPE == "patch" ]]
+          then
+              NEXT_VERSION=v$(./.github/workflows/util/semver bump $RELEASE_TYPE $RELEASE_VERSION)
```
israel-hdez, line 71: Wouldn't it better to simply use this line replacing the whole if/else?

And I echo that I wanted to remove the dependency on that `semver` script :-(

.github/workflows/release.yml
```diff
@@ -0,0 +1,203 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.inputs.repository }}
+          ref: ${{ github.event.inputs.releasing_branch }} 
+        
+      - name: Configure git
+        run: |
+          git config user.email 'kiali-dev@googlegroups.com'
+          
+          git config user.name 'kiali-bot'
```
israel-hdez, line 116: Same comment as in server PR: Probably this step is not needed, because actions may already have something configured that can be used.

.github/workflows/release.yml
```diff
@@ -0,0 +1,203 @@
...
+        run: git checkout master
+      
+      # Anticipated preparation of Makefile so that it contains the released
+      # version when creating the git tag.
+      #
+      # This change to the Makefile is not done for patch releases, because that
+      # would break the next minor build (remember we have already switched to master).
+      - name: Prepare Makefile
+        if: ${{ github.event.inputs.release_type != 'patch'}}
+        run: sed -i -r 's/^VERSION \?= v.*/VERSION \?= v$RELEASE_VERSION/' Makefile
```
israel-hdez, line 133: Doesn't `RELEASE_VERSION` already contains the `v`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,203 @@
...
+            
+            git commit -m "Prepare for next version"
+
+            git push origin $(git rev-parse HEAD):refs/heads/$BUILD_TAG
+
+            echo "{\"title\": \"Prepare for next version\", \"body\": \"Please, merge to update version numbers and prepare for release $NEXT_VERSION.\", \"head\": \"kiali-bot:$BUILD_TAG\", \"base\": \"master\"}"
+            curl -H "Authorization: token ${{ secrets.GH_TOKEN }}" \
+              -H "Content-Type: application/json" \
+              -d "{\"title\": \"Prepare for next version\", \"body\": \"Please, merge to update version numbers and prepare for release $NEXT_VERSION.\", \"head\": \"kiali-bot:$BUILD_TAG\", \"base\": \"master\"}" \
+              -X POST $GH_PULL_URI
```
israel-hdez, line 175: Like in the server side, perhaps use `gh` CLI for better clarity?

israel-hdez (commented):
Minor comments

.github/workflows/release.yml
leandroberetta: I didn't know you wanted to remove, I was thinking on removing it a few days ago, but finally didn't do it, well... until now, what do you think of the following:

https://github.com/kiali/kiali/pull/4980/files#diff-4d71906b252fbf817a4b47220031fb7a35caaeb33785596cc8e5b0b392357dd3

It is based on the groovy function "bump" that the operator is currently using.

leandroberetta: I replaced the semver file with this new approach on the server release pipeline, I'm thinking on doing the same thing on the rest of the pipelines, is that ok for you?.

.github/workflows/release.yml
leandroberetta: Yes, probably this is wrong, I'll fix this soon.

.github/workflows/release.yml
leandroberetta: Sure, I've already updated the server release pipeline to use gh cli. The rest of the pipelines will be modified soon.

.github/workflows/release.yml
leandroberetta: It is needed:

Author identity unknown

*** Please tell me who you are.

I searched a little bit but I only found workarounds for this.

jmazzitelli:
Does that 1.50.0-master represent what `master` would look like when all done?

leandroberetta:
> Does that 1.50.0-master represent what `master` would look like when all done?

Yes, with the updated "docs" folder containing the releases.

jmazzitelli (approved):

PR #138, opened 2022-05-09
Add cert-manager support for kiali-server ingress

cbarton:
Create the `Certificate` resource and mounts that certificate into the
pod for use according to the following settings:

```yaml
identity:
  cert_manager:
    enabled: true
    issuer:
      name: <issuer-name>
      kind: ClusterIssuer # Default is Issuer
```


pcrummett (approved):

jmazzitelli:
Can you create a github issue with the enhancement you are asking for? Without an issue this PR is addressing, we can't judge what it is or its usefulness to the community.

Is this an OpenShift or non-OpenShift enhancement?  What is this doing/what is the purpose?

Also, this PR only affects the server helm chart - I don't see a link to an operator PR for operator changes as well. We will need to incorporate the same functionality in the operator (the operator is the main/recommended installation mechanism for Kiali - the server helm chart is the secondary alternative mechanism). Please see the checklists: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#how-to-change-resources-or-configuration

Finally - please provide some test procedures. Manual test steps at minimum, or better yet, if possible, add an automated [molecule](https://github.com/kiali/kiali-operator/tree/master/molecule) test.

jmazzitelli (changes requested):
some additional things are needed - see my prior comment

kiali-server/templates/_helpers.tpl
```diff
@@ -84,14 +84,27 @@ Determine the default web root.
...
+{{/*
+Determine if using a built in or autogenerated certificate.
+*/}}
+{{- define "kiali-server.identity.provider" }}
+  {{- if .Capabilities.APIVersions.Has "route.openshift.io/v1" }}
+    openshift/v1
+  {{- else if .Values.identity.cert_manager.enabled }}
+    cert_manager
+  {{- else }}
+  ""
```
jmazzitelli, line 96: trivial - indent off
```suggestion
    ""
```

kiali-server/templates/certificate.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: {{ include "kiali-server.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  secretName: {{ include "kiali-server.fullname" . }}-cert-secret
+  secretTemplate:
+    labels:
+      {{- include "kiali-server.labels" . | nindent 8 }}
+  dnsNames:
+    - {{ required ".Values.server.web_fqdn is required when .Values.identity.cert_manager is enabled" .Values.server.web_fqdn }}
```
jmazzitelli, line 14: trivial - kiali indentation standard is not to indent list items. leave it at same level as parent
```suggestion
  - {{ required ".Values.server.web_fqdn is required when .Values.identity.cert_manager is enabled" .Values.server.web_fqdn }}
```

kiali-server/templates/service.yaml
```diff
@@ -31,11 +31,21 @@ spec:
...
+    {{- if .Values.deployment.service_port }}
+    targetPort: {{ .Values.server.port }}
+    port: {{ .Values.deployment.service_port }}
+    {{- else }}
     port: {{ .Values.server.port }}
+    {{- end }}
   {{- if .Values.server.metrics_enabled }}
   - name: http-metrics
     protocol: TCP
+    {{- if .Values.deployment.service_metrics_port }}
```
jmazzitelli, line 43: what is "deployment.service_metrics_port"? That is not an existing setting correct? See dev checklist when adding new settings.

kiali-server/templates/service.yaml
```diff
@@ -31,11 +31,21 @@ spec:
...
+    targetPort: {{ .Values.server.metrics_port }}
+    port: {{ .Values.deployment.service_metrics_port }}
+    {{- else }}
     port: {{ .Values.server.metrics_port }}
+    {{- end }}
```
jmazzitelli, line 48: changes to this service.yaml template does not appear to be related to the issue of certificate creation. If so, this should be in a separate PR under a separate enhancement github issue.

jmazzitelli (changes requested):
some additional things.

This looks like a less-than-trivial change (in fact, it looks like two separate enhancement requests).

Github issues need to be created where discussions and descriptions of the issue can take place.

In addition, it appears some new settings are being introduced. These need to be added to the Kiali CR schema and its docs in addition to other places (particularly in the operator).

See the dev checklist for the things that need to be done before these changes can be incorporated.

cbarton:
@jmazzitelli Thanks for the quick review. Upon a further discussion internally, we can get cert-manager support by leveraging the `deployment.custom_secrets` instead of adding the burden for review/development on the operator for first class support. You are correct that this includes a change to allow for an override on the Service ports, so I'll open an issue on kiali/kiali for that discussion.

PR #139, opened 2022-05-11
Helm charts release schedule

leandroberetta:
This PR adds the scheduling mechanism on Mondays.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,9 @@
 name: Release
 
 on:
-  # schedule:
-  #   # Every Friday at 19:00 (UTC)
-  #   - cron: '00 19 * * FRI'
+  schedule:
+    # Every Friday at 19:00 (UTC)
+    - cron: '00 19 * * FRI'
   workflow_dispatch:
     inputs:
       release_type:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -78,7 +78,7 @@ jobs:
         if [ -z ${{ github.event.inputs.release_type }} ];
         then
           DO_RELEASE=$(python minor.py)
-          if [[ $DO_RELEASE == "0" ]]
+          if [[ $DO_RELEASE == "1" ]]
           then
             echo "::set-output name=release_type::minor"
           else
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,14 @@
 name: Release
 
 on:
+  # schedule:
+  #   # Every Friday at 19:00 (UTC)
+  #   - cron: '00 19 * * FRI'
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -21,53 +23,111 @@ jobs:
...
         then
-            RELEASE_VERSION=$(./.github/workflows/util/bump.py $RELEASE_TYPE $RELEASE_VERSION)
+          RELEASE_VERSION=$(python bump.py $RELEASE_TYPE $RELEASE_VERSION)
+        elif [[ $RELEASE_TYPE == "minor" ]]
+        then
+          RELEASE_VERSION=$RELEASE_VERSION
         fi
 
-        echo "::set-output name=release_version::$RELEASE_VERSION"      
+        echo "::set-output name=release_version::$RELEASE_VERSION"
```
jmazzitelli, line 110: does this have spaces at the end of line? I see alot of changes recently removing end of line spaces - I assume you missed this one.

.github/workflows/release.yml
```diff
@@ -1,12 +1,14 @@
 name: Release
 
 on:
+  schedule:
+    # Every Friday at 19:00 (UTC)
+    - cron: '00 19 * * FRI'
```
jmazzitelli, line 6: I'm confused with this PR - is this releasing every week via this cron schedule? I'm sure that isn't what its doing, but without comments in here explaining, it is confusing.

jmazzitelli (changes requested):

.github/workflows/release.yml
leandroberetta: I missed that one, yes.

.github/workflows/release.yml
leandroberetta: This PR adds the schedulling mechanism to the Helm charts release pipeline. I'm not sure when you say that it isn't what is doing. 

This PR adds the cron, the release type selection (checking if it needs to skip the release if we are not in the end sprint week) and also adds some checks similar to the server and operator ones.

leandroberetta: I added the "do not merge" label to wait after the next release that will be manual.

jmazzitelli: I thought we did this every 3 weeks. This says in the comment it does this every Friday. So I guess it determines later if it is in the third week to actually do the work (the other 2 weeks it just does nothing?)

leandroberetta: Yes, with a cron sentence you can't fire every 3 weeks, it was commented on the following PR too: https://github.com/kiali/kiali/pull/5037 (comment #2).

So, I don't like the fact that we will see a release pipeline run for nothing, but it's the way GH Actions works today.

jmazzitelli (approved):

PR #141, opened 2022-05-25
[helm] Update the HPA api version to be used by default in the server helm chart

jmazzitelli:
autoscaling API version `v2beta2` is being removed in newer clusters. Default the server helm chart to `v2`.

If you are running an older cluster that does not have `v2`, simply set this value to the `v2beta2` version at the same time you are define the `spec` for the HPA.

Part of: https://github.com/kiali/kiali/issues/5022

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -39,7 +39,7 @@ deployment:
   custom_secrets: []
   host_aliases: []
   hpa:
-    api_version: "autoscaling/v2beta2"
+    api_version: "autoscaling/v2"
     spec: {}
   image_digest: "" # use "sha256" if image_version is a sha256 hash (do NOT prefix this value with a "@")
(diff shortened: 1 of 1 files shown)


jmazzitelli:
installed successfully both without HPA and with an HPA:

With HPA:

```
helm install --namespace istio-system --set auth.strategy="anonymous" --set deployment.image_version=v1.51 --set deployment.hpa.spec.minReplicas=1 --set deployment.hpa.spec.maxReplicas=2 kiali-server _output/charts/kiali-server-1.52.0-SNAPSHOT.tgz 
```

Results in:

```
$ kubectl get horizontalpodautoscalers.autoscaling -n istio-system -o name
horizontalpodautoscaler.autoscaling/kiali
```

jshaughn (approved):
LOK

PR #146, opened 2022-06-09
Patch release create PR targetting master now

leandroberetta:
For patch releases, the PR target should be master, not the release branch (ie v1.50).

The following PR is a test on my fork, seems to be correct:

https://github.com/leandroberetta/helm-charts/pull/10/files


jmazzitelli (approved):
> The following PR is a test on my fork, seems to be correct:

Yes, the /docs content should be changed on the master branch, as it is there in your test fork.

PR #155, opened 2022-07-08
feat: allow custom annotations on kiali cr

oreonl:
This would allow specifying custom annotations on the kiali cr when it is created by the chart. I specifically need this so that I can tell Argo CD to have it in a later sync wave so during a delete operation the cr is removed before the operator.

(updated by @jmazzitelli July 8, 2022)
fixes: https://github.com/kiali/kiali/issues/5297

### DIFF
--- a/kiali-operator/templates/kiali-cr.yaml
+++ b/kiali-operator/templates/kiali-cr.yaml
@@ -11,11 +11,11 @@ metadata:
   name: {{ .Values.cr.name }}
   labels:
   {{- include "kiali-operator.labels" . | nindent 4 }}
-annotations:
-  ansible.sdk.operatorframework.io/verbosity: {{ .Values.debug.verbosity | quote }}
-  {{- if .Values.cr.annotations }}
-  {{- toYaml .Values.cr.annotations | nindent 2 }}
-  {{- end }}
+  annotations:
+    ansible.sdk.operatorframework.io/verbosity: {{ .Values.debug.verbosity | quote }}
+    {{- if .Values.cr.annotations }}
+    {{- toYaml .Values.cr.annotations | nindent 4 }}
+    {{- end }}
 spec:
   {{- toYaml .Values.cr.spec | nindent 2 }}
 ...
--- a/kiali-operator/templates/kiali-cr.yaml
+++ b/kiali-operator/templates/kiali-cr.yaml
@@ -13,6 +13,9 @@ metadata:
   {{- include "kiali-operator.labels" . | nindent 4 }}
 annotations:
   ansible.sdk.operatorframework.io/verbosity: {{ .Values.debug.verbosity | quote }}
+  {{- if .Values.cr.annotations }}
+  {{- toYaml .Values.cr.annotations | nindent 2 }}
+  {{- end }}
 spec:
   {{- toYaml .Values.cr.spec | nindent 2 }}
 ...
--- a/kiali-operator/values.yaml
+++ b/kiali-operator/values.yaml
@@ -77,9 +77,10 @@ cr:
   # then this is the namespace where the CR will be created (the default will be the operator namespace).
   namespace: ""
 
+  # Annotations to place in the Kiali CR metadata.
+  annotations: {}
+
   spec:
     deployment:
       accessible_namespaces:
       - '**'
-
-  annotations: {}
--- a/kiali-operator/values.yaml
+++ b/kiali-operator/values.yaml
@@ -81,3 +81,5 @@ cr:
     deployment:
       accessible_namespaces:
       - '**'
+
+  annotations: {}


kiali-operator/values.yaml
```diff
@@ -81,3 +81,5 @@ cr:
     deployment:
       accessible_namespaces:
       - '**'
+
+  annotations: {}
```
jmazzitelli, line 85: Please move this up to line 79 (place it above `spec`) so it doesn't confuse people thinking this is under "spec".
Also, please add a comment to `annotations: {}` to explain this to users - use it as a kind of documentation.

Something like:
```
  ...
  namespace: ""

  # Annotations to place in the Kiali CR metadata. 
  annotations: {}

  spec:
    ...
```

jmazzitelli (changes requested):

kiali-operator/values.yaml
oreonl: Made the requested changes.

jmazzitelli:
Well, this is embarrassing :)

I think you uncovered a bug. We should fix this in this PR.

I just ran this:

```
make build-helm-charts
helm install --set cr.annotations.foo=bar \
   --set cr.create=true \
   --set cr.namespace=istio-system \
   --namespace kiali-operator \
   --create-namespace \
   kiali-operator \
   _output/charts/kiali-operator-1.54.0-SNAPSHOT.tgz
```

And take a look at the Kiali CR that is generated:

```
$ kubectl get kiali kiali -n istio-system -o yaml
annotations:
  ansible.sdk.operatorframework.io/verbosity: "1"
  foo: bar
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  annotations:
    meta.helm.sh/release-name: kiali-operator
    meta.helm.sh/release-namespace: kiali-operator
...
```

So apparently, the Kiali helm chart is putting the `annotations` section at the top root - rather than under `metadata`.

We need to fix this.

Can you move the `annotations` section under `metadata` in the template?

kiali-operator/templates/kiali-cr.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
   {{- include "kiali-operator.labels" . | nindent 4 }}
 annotations:
   ansible.sdk.operatorframework.io/verbosity: {{ .Values.debug.verbosity | quote }}
+  {{- if .Values.cr.annotations }}
+  {{- toYaml .Values.cr.annotations | nindent 2 }}
+  {{- end }}
```
jmazzitelli, line 18:
```suggestion
  annotations:
    ansible.sdk.operatorframework.io/verbosity: {{ .Values.debug.verbosity | quote }}
    {{- if .Values.cr.annotations }}
    {{- toYaml .Values.cr.annotations | nindent 4 }}
    {{- end }}
```

jmazzitelli (changes requested):
indent the `annotations` section so it is under `metadata`

kiali-operator/templates/kiali-cr.yaml
oreonl: Ha. I actually had a problem when I first added the the `cr.annotations` because I was indenting 4 and it was giving me an error. I realized it was only indenting 2 but totally missed that it was because it wasn't under `metadata`.

jmazzitelli:
Now running that command I describe above works - notice the annotations are correctly placed in the metadata:

```
$ kubectl get kiali kiali -n istio-system -o yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  annotations:
    ansible.sdk.operatorframework.io/verbosity: "1"
    foo: bar
    meta.helm.sh/release-name: kiali-operator
    meta.helm.sh/release-namespace: kiali-operator
  creationTimestamp: "2022-07-09T00:55:33Z"
  generation: 1
  labels:
    app: kiali-operator
    app.kubernetes.io/instance: kiali-operator
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: kiali-operator
    app.kubernetes.io/part-of: kiali-operator
    app.kubernetes.io/version: v1.54.0-SNAPSHOT
    helm.sh/chart: kiali-operator-1.54.0-SNAPSHOT
    version: v1.54.0-SNAPSHOT
  name: kiali
  namespace: istio-system
  resourceVersion: "363695"
  uid: e098fdee-6cc2-4286-92c8-d9ca3c419c32
spec:
  deployment:
    accessible_namespaces:
    - '**'
```

jmazzitelli:
I also tested the typical use case of not specifying custom annotations and it all works as before - so this is backwards compat.

jmazzitelli (approved):

stevehipwell:
@jmazzitelli when will this be released?

israel-hdez:
> @jmazzitelli when will this be released?

Monday 25th

PR #159, opened 2022-08-18
move crd into templates to ensure it's installed with chart

lwabish:
currently the crd is out of the templates directory, so by default it's not installed.
kaili operator can't even work without the CRD.
maybe we could put the crd into templates directory?
Signed-off-by: lwabish <wubw@pku.edu.cn>


jmazzitelli:
Incorrect. This is the completely wrong way to install CRDs using Helm. Please read [the helm developer documentation when dealing with CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/).

Before submitting PRs, please submit a github discussion or issue and ask how something can be or should be done in Kiali. It will help avoid you wasting time creating PRs that are incorrect or will not be accepted.

lwabish:
Thanks for the link, didn't know about helm chart crd directory.
But I did run into a situation that the chart is installed but the CRD is not installed.
I opened a discussion here and will do some more digging into this : #5398

PR #167, opened 2022-09-13
add kiali server deployment annotations

lamc:
contributing to kiali-server helm chart to add deployment annotations to template. 

we have a use case where use deployment annotations to mass target within our automation.


jmazzitelli:
If you would like a feature like this added to Kiali, please submit a github issue first so it can be discussed with the community. Do not just submit a PR as it will likely not be accepted without a github issue in place with a discussion of the enhancement request.

In this case, I would have indicated that there is more to this than just adding to the server helm chart. Please see the developer checklist when you want to add a server config like this: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting

A github issue will be required so we can track the PRs that will be required in at least the helm chart repo and the operator repo.

lamc:
okay understood. I will close this PR and follow the contribution docs

PR #176, opened 2022-10-22
Fix for Apple MacOS M1

acierto:
fixes: https://github.com/kiali/kiali/issues/5580

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ HELM_IMAGE_REPO_SERVER ?= quay.io/kiali/kiali
 DORP ?= docker
 
 # When building the helm chart, this is the helm version to use
-HELM_VERSION ?= v3.5.2
+HELM_VERSION ?= v3.10.1
 
 .PHONY: help
 help: Makefile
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ HELM_IMAGE_REPO_SERVER ?= quay.io/kiali/kiali
 DORP ?= docker
 
 # When building the helm chart, this is the helm version to use
-HELM_VERSION ?= v3.5.2
+HELM_VERSION ?= v3.10.1
```
jmazzitelli, line 21: is this bump of helm required for ARM support? If not, it would not be related to this issue and so should not be included.

jmazzitelli:
Please create a Kiali issue and associate it with this PR (i.e. put `fixes: <link to issue>` in the description of this PR). Please do not submit PRs to the Kiali project without an associated issue so we can track issues better. Thanks.

Putting on a DNM label until we get an issue for this PR.

Makefile
acierto: Yeah it is, as ARM support started from 3.6.0. and if to bump it, why not to bump to the latest.
Let me know what do you think

acierto:
> Please create a Kiali issue and associate it with this PR (i.e. put `fixes: <link to issue>` in the description of this PR). Please do not submit PRs to the Kiali project without an associated issue so we can track issues better. Thanks.
> 
> Putting on a DNM label until we get an issue for this PR.

Thank you for the comments. Created an issue and associated it with this PR

jmazzitelli:
Works on my machine. Confirmed it pulls down helm 3.10.1, and the build-helm-charts target still works. Nothing breaks. I don't have an ARM machine to test, but I assume that will work since that's the whole point :)

jmazzitelli (approved):

PR #180, opened 2022-11-22
kiali SA to be given read-only perms unless using anonymous strategy with view_only_mode off

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/5611


leandroberetta (approved):
Verified. LGTM.

PR #189, opened 2023-01-23
remove list secrets perm, mount cluster secrets to file system, support autodiscovery of cluster secrets

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/5738

### DIFF
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -171,3 +171,26 @@ Determine the root namespace - default is where Kiali is installed.
   {{- .Release.Namespace }}
 {{- end }}
 {{- end }}
+
+{{/*
+Autodetect remote cluster secrets if enabled - looks for secrets in the same namespace where Kiali is installed.
+Returns a JSON dict whose keys are the cluster names and values are the cluster secret data.
+*/}}
+{{- define "kiali-server.remote-cluster-secrets" -}}
+{{- $theDict := dict }}
+{{- if .Values.kiali_feature_flags.clustering.autodetect_secrets.enabled }}
+  {{- $secretLabelToLookFor := (regexSplit "=" .Values.kiali_feature_flags.clustering.autodetect_secrets.label 2) }}
+  {{- $secretLabelNameToLookFor := first $secretLabelToLookFor }}
+  {{- $secretLabelValueToLookFor := last $secretLabelToLookFor }}
+  {{- range $i, $secret := (lookup "v1" "Secret" .Release.Namespace "").items }}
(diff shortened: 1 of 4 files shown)


jmazzitelli:
`helm template --show-only templates/deployment.yaml kiali-server _output/charts/kiali-server*.tgz` will show the new secret mounts when its implemented.

jmazzitelli:
Here's a test to see the server helm chart with the new secret mounts.
Run this:

```
make build-helm-charts && helm template --show-only templates/deployment.yaml --set kiali_feature_flags.clustering.clusters[0].name=foo --set kiali_feature_flags.clustering.clusters[0].secret_name=foosecret kiali-server _output/charts/kiali-server*.tgz
```

and the output should show this:

```yaml
        volumeMounts:
...
        - name: foo
          mountPath: "/kiali-remote-cluster-secrets/foo"
...
      volumes:
...
      - name: foo
        secret:
          secretName: foosecret
...
```

jmazzitelli:
The server helm chart now supports autodiscovery of remote cluster secrets. This requires the user running helm install to have permission to read secrets in the deployment namespace.

To test that, first make sure you build everything: `make build-helm-charts` and then:

1. Don't install any remote secrets and see that it works as-is:

```
helm install --debug --namespace istio-system kiali-server _output/charts/kiali-server*.tgz
```

Look at the resulting deployment and you should not see any remote cluster secrets mounted.

2. Install a remote secret (this assumes you have a "minikube" context): `istioctl x create-remote-secret --name my-remote-cluster-name --context minikube --namespace istio-system | kubectl apply -f -`

Now install the server and see the secret mounted in the deployment:

```
helm install --debug --namespace istio-system kiali-server _output/charts/kiali-server*.tgz
...
        - name: my-cluster-name
          mountPath: "/kiali-remote-cluster-secrets/my-cluster-name"
...
      - name: my-cluster-name
        secret:
          secretName: istio-remote-secret-my-cluster-name
...
```

3. Install a second secret (call the cluster "my2-cluster-name") and re-run the test you did in step 2 only now confirm you can see two secrets mounted.

4. Disable autodiscovery and see that no secrets are mounted (the output should look like step 1 - there are no secret volumes declared in the deployment):

```
helm install --debug --namespace istio-system --set kiali_feature_flags.clustering.autodetect_secrets.enabled=false kiali-server _output/charts/kiali-server*.tgz
```

jmazzitelli:
Using the helm install above, you can quickly test the server-side PR (i.e. make sure the server can read and process those secrets mounted on the file system). So install the server via the server helm chart and then just test the server.

leandroberetta (approved):
Looks good, works well too, tested the different scenarios.

PR #193, opened 2023-02-01
Always use Kiali's Helm binary

leandroberetta:
For https://github.com/kiali/kiali/pull/5783


jmazzitelli (approved):
i didn't run, but it looks like it does what we want.

PR #203, opened 2023-03-22
Update Kiali server values files with example for components and their namespaces

lisenet:
Closes https://github.com/kiali/kiali/issues/5935

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -82,6 +82,17 @@ external_services:
     enabled: true
   istio:
     root_namespace: ""
+#    component_status:
+#      enabled: true
+#      components:
+#      - app_label: istiod
+#        is_core: true
+#      - app_label: istio-ingressgateway
+#        is_core: true
+#        namespace: istio-ingress
+#      - app_label: istio-egressgateway
+#        is_core: false
+#        namespace: istio-egress
 
 identity: {}
   #cert_file:


jmazzitelli (commented):
I have thought about this, and I've now come to the conclusion that I don't think we want this. This change is just proliferating/duplicating documentation that we already have - it is now one more place we have to maintain the docs for this. We already document these settings elsewhere: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.istio.component_status.enabled

I realize the user in this case said it was hard to find, but honestly, just go to kiali.io and the docs are there. Let's not duplicate docs where we now have multiple places that we have to touch when docs have to change or be maintained. Also, what would then stop anyone from adding commented docs for EVERYTHING in the config? Soon this will grow to essentially be a duplication of the same full docs we already have.

We could make it a little more clear here that "To see what settings are available, see the Kiali CR reference documention at kiali.io". Indeed, we already infer this inside the [values.yaml already](https://github.com/kiali/helm-charts/blob/v1.66.0/kiali-server/values.yaml#L17) when it says, `# Settings that mimic the Kiali CR which are placed in the ConfigMap.`. We should just make it more clear and provide a URL link to that documentation, rather than just duplicate existing docs.

So, I'm -1 on this. I'll ask @jshaughn what he thinks about this.

lisenet:
@jmazzitelli thanks, appreciate your response. Just to check something, you do have commented out sections added to the [values.yaml file](https://github.com/kiali/helm-charts/blob/v1.66.0/kiali-server/values.yaml) already, is there a policy somewhere of what commented out config sections can and cannot be added in the future?

jmazzitelli:
> is there a policy somewhere of what commented out config sections can and cannot be added in the future?

Yes, it [mentions this in that values.yaml file](https://github.com/kiali/helm-charts/blob/v1.66.0/kiali-server/values.yaml#L18) where it says `Note that only those values used by the Helm Chart will be here.`

So if the helm charts themselves need the settings to do what they need to do, you will see them in the yaml file. That's why only a subset is specified - only those settings that are necessary and needed by the helm chart itself are in values.yaml.

If the value is commented out, it will use the documented default value. For example, [ingress.enabled is commented out](https://github.com/kiali/helm-charts/blob/v1.66.0/kiali-server/values.yaml#L52), but [its default is determined](https://github.com/kiali/helm-charts/blob/v1.66.0/kiali-server/templates/_helpers.tpl#L117-L130) in _helpers.tpl and is used by the [ingress template here](https://github.com/kiali/helm-charts/blob/v1.66.0/kiali-server/templates/ingress.yaml#L2) - that's why you see it in values.yaml.

PR #207, opened 2023-04-28
provide permission to support override_yaml.spec.host on openshift

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/6083

### DIFF
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -148,6 +148,7 @@ rules:
 - apiGroups: ["route.openshift.io"]
   resources:
   - routes
+  - routes/custom-host
   verbs:
   - create
   - delete


jmazzitelli:
How to test:

1. Log into an OpenShift cluster that has Istio installed in the namespace `istio-system`.
2. cd to your local kiali helm-charts repo, pull this PR, and then build the helm charts via `make clean build-helm-charts`
3. Install the operator and a Kiali CR via helm using this command (notice this command will override the ingress yaml with the spec.host setting that requires the new permission, which is what this PR adds):
```
helm install -n kiali-operator --create-namespace --set image.tag=latest \
  --set cr.create=true \
  --set cr.spec.deployment.namespace=istio-system \
  --set cr.spec.deployment.ingress.override_yaml.spec.host=my-custom-kiali-host.local \
  --set cr.spec.deployment.ingress.override_yaml.spec.to.name=kiali \
  kiali-operator _output/charts/kiali-operator-*.tgz
```
4. Wait for Kiali to be installed, specifically wait for the Kiali route. When the route is up, verify the `spec.host` has the custom hostname of `my-custom-kiali-host.local`:
```
$ oc get route kiali -n istio-system -o jsonpath='{.spec.host}{"\n"}'
my-custom-kiali-host.local
```
5. When you are finished, uninstall everything to clean up:
```
oc delete kiali kiali -n kiali-operator && helm uninstall kiali-operator -n kiali-operator && oc delete ns kiali-operator
```

jmazzitelli:
You can see what the bug looks like if you run this command (this just installs the 1.67 operator helm chart with the previous comment's settings - so you can see what happens if you try to override the route's spec.host setting):

```
helm install -n kiali-operator --create-namespace --set cr.create=true  \
  --set cr.spec.deployment.namespace=istio-system  \
  --set cr.spec.deployment.ingress.override_yaml.spec.host=my-custom-kiali-host.local  \
  --set cr.spec.deployment.ingress.override_yaml.spec.to.name=kiali \
  --version 1.67.0 kiali-operator kiali/kiali-operator
```

This results in an error you can see in the Kiali CR's status field:

```
$ oc get kiali kiali -n kiali-operator -o yaml
...
    message: 'Route kiali: Failed to create object: b''{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Route.route.openshift.io
      \\"kiali\\" is invalid: spec.host: Forbidden: you do not have permission to
      set the host field of the route","reason":"Invalid","details":{"name":"kiali","group":"route.openshift.io","kind":"Route","causes":[{"reason":"FieldValueForbidden","message":"Forbidden:
      you do not have permission to set the host field of the route","field":"spec.host"}]},"code":422}\n'''
    reason: Failed
    status: "False"
    type: Failure
...
```

If you try this, clean up everything via:

```
oc delete kiali kiali -n kiali-operator && helm uninstall kiali-operator -n kiali-operator && oc delete ns kiali-operator
```

PR #212, opened 2023-05-23
set oauth redirect to route with non-standard port

jmazzitelli:
part of https://github.com/kiali/kiali/issues/6180

### DIFF
--- a/kiali-server/templates/oauth.yaml
+++ b/kiali-server/templates/oauth.yaml
@@ -10,6 +10,9 @@ metadata:
     {{- include "kiali-server.labels" . | nindent 4 }}
 redirectURIs:
 - {{ .Values.kiali_route_url }}
+{{- if .Values.server.web_port }}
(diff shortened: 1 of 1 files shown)


jmazzitelli:
to test install on OpenShift, specifying a server.web_port. Note that to install on OpenShift using the server helm chart, you need to know what the kiali route will be beforehand - you pass it via `--set kiali_route_url`:

```
helm install -n istio-system --set kiali_route_url=https://kiali-istio-system.apps-crc.testing --set server.web_port=31102 kiali-server _output/charts/kiali-server-*-SNAPSHOT.tgz
```

Now check the redirect URIs - ensure they are correct:

```
$ oc get oauthclient kiali-istio-system -o jsonpath={.redirectURIs} | jq
[
  "https://kiali-istio-system.apps-crc.testing",
  "https://kiali-istio-system.apps-crc.testing:31102"
]
```

jshaughn (approved):
I tested this against 1.65 (ossm 2.4) and it is working.

PR #219, opened 2023-08-07
Added default value for kiali_urls.

hhovsepy:
Operator PR https://github.com/kiali/kiali-operator/pull/672


jmazzitelli (changes requested):
@hhovsepy you don't need this PR. These settings you see in `values.yaml` only need to be here IF the helm charts themselves use the values. See the comment here: https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L18

Because nothing in the helm charts themselves need this new value, you don't have to put this here. You can close the PR. No need to merge this.

PR #221, opened 2023-08-07
fix icon link in the helm charts metadata

jmazzitelli:
we've been releasing the helm charts with a broken icon. This is a minor issue, but would be nice if we publish future charts that point to an existing icon.


aljesusg (approved):
LGFM

PR #236, opened 2023-12-01
Added additional_display_details to be default set.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6665
Original feature: https://github.com/kiali/kiali/issues/1567

Make sure that in helm installation the "additional_display_details" a set, similar to operator one: https://github.com/kiali/kiali-operator/blob/master/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L8

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -18,6 +18,11 @@ kiali_route_url: ""
 # Note that only those values used by the Helm Chart will be here.
 #
 
+additional_display_details:
+- annotation: kiali.io/api-spec
+  icon_annotation: kiali.io/api-type
+  title: API Documentation
+
 istio_namespace: "" # default is where Kiali is installed
 
 auth:


kiali-server/values.yaml
```diff
@@ -18,6 +18,11 @@ kiali_route_url: ""
 # Note that only those values used by the Helm Chart will be here.
 #
 
+additional_display_details:
+- annotation: kiali.io/api-spec
+  icon_annotation: kiali.io/api-type
+  title: API Documentation
+
```
jmazzitelli, line 25: Is this really needed? Note the comment above that says, `only those values used by the Helm Chart will be here`. We only put stuff in here that the actual Helm charts reference and use. We don't set simple defaults here as there should be no need.

jmazzitelli: Hmm... I guess this isn't a hardcoded default in the server? I'm trying to think if we really want this in the values file or not - we never put defaults in here unless the charts actually use them, but in this case I don't see a way around having to put this in here.

jmazzitelli: OK, I think the difference here is that this setting is an array of items. So if we hardcode this default, but the user sets his own, that default list item will be overwritten. The user will need to know to add it back in if they want to see the Kiali API stuff.  So, yes, I can see why we could put this in there, because it expicitly shows the user the kiali API spec list item and if they want their own, they can just add it to the list if they want to keep the kiali one.

jmazzitelli (approved):

kiali-server/values.yaml
hhovsepy: right, this is not hardcoded values but says where to look the values

PR #242, opened 2024-01-11
Add top level `clustering` field

nrfox:
Adds a top level `clustering` field to the kiali spec.

Relates to kiali/kiali#6243

### DIFF
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -178,8 +178,8 @@ Returns a JSON dict whose keys are the cluster names and values are the cluster 
 */}}
 {{- define "kiali-server.remote-cluster-secrets" -}}
 {{- $theDict := dict }}
-{{- if .Values.kiali_feature_flags.clustering.autodetect_secrets.enabled }}
-  {{- $secretLabelToLookFor := (regexSplit "=" .Values.kiali_feature_flags.clustering.autodetect_secrets.label 2) }}
+{{- if .Values.clustering.autodetect_secrets.enabled }}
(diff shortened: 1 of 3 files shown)


jmazzitelli (changes requested):
Need to do 2 things:

1. You should remove [the deprecated feature flags from values.yaml](https://github.com/kiali/helm-charts/blob/v1.78.0/kiali-server/values.yaml#L101-L105).

2. Then you need to change the templates to use these new settings:

So you will want to change this stuff:
* https://github.com/kiali/helm-charts/blob/v1.78.0/kiali-server/templates/_helpers.tpl#L181-L182
* https://github.com/kiali/helm-charts/blob/v1.78.0/kiali-server/templates/deployment.yaml#L134
* https://github.com/kiali/helm-charts/blob/v1.78.0/kiali-server/templates/deployment.yaml#L177

You can see all these places that use the feature flags via:
```sh
$ grep -R "kiali_feature_flags.clustering" *
templates/_helpers.tpl:{{- if .Values.kiali_feature_flags.clustering.autodetect_secrets.enabled }}
templates/_helpers.tpl:  {{- $secretLabelToLookFor := (regexSplit "=" .Values.kiali_feature_flags.clustering.autodetect_secrets.label 2) }}
templates/deployment.yaml:        {{- range .Values.kiali_feature_flags.clustering.clusters }}
templates/deployment.yaml:      {{- range .Values.kiali_feature_flags.clustering.clusters }}
```

This is the server helm chart (provided for convienence) so we are less strict about backwards compatibility. We will, however, make a note of this in the release notes so people are aware of this change. cc @jshaughn  (he produces the release notes).

jmazzitelli (approved):

PR #246, opened 2024-02-12
Add git config back

leandroberetta:
We need to add some git configuration back into the pipeline because is required to push or create tags. Current releases failed because of this: https://github.com/kiali/helm-charts/actions/runs/7868923370 

For reference, this is the PR that made the change: https://github.com/kiali/helm-charts/pull/243

For now, until we investigate a way of removing this configuration, we will need it for release.


josunect (approved):
I think 2FA should not be affected for this? 

LGTM

PR #248, opened 2024-02-14
feat(kiali-server): add custom nodeport for service

emreberber:
Related Issue ➜ [Kiali Server Helm Chart Support Custom NodePort](https://github.com/kiali/kiali/issues/7093)

When Kiali Server is deployed to hundreds of Kubernetes clusters, it would be very good to have this feature to access all clusters with the same `nodePort`. This PR was prepared to use a custom port for `NodePort`. 

In Values, we can get the nodePort value as follows:

```
server:
  port: 20001
  nodePort: 32475
  observability:
    metrics:
      enabled: true
      port: 9090
      nodePort: 32476
  web_root: ""
```

We can use it in the Service as follows:

```
{{- if eq .Values.deployment.service_type nodePort }}
    nodePort: {{ .Values.server.nodePort }}
{{- end }}
```

#7093

### DIFF
--- a/kiali-server/templates/service.yaml
+++ b/kiali-server/templates/service.yaml
@@ -34,7 +34,7 @@ spec:
   {{- end }}
     protocol: TCP
     port: {{ .Values.server.port }}
-    {{- if eq .Values.deployment.service_type "NodePort" }}
+    {{- if and (not (empty .Values.server.node_port)) (eq .Values.deployment.service_type "NodePort") }}
     nodePort: {{ .Values.server.node_port }}
     {{- end }}
   {{- if .Values.server.observability.metrics.enabled }}
--- a/kiali-server/templates/service.yaml
+++ b/kiali-server/templates/service.yaml
@@ -42,9 +42,6 @@ spec:
     appProtocol: http
     protocol: TCP
     port: {{ .Values.server.observability.metrics.port }}
-    {{- if eq .Values.deployment.service_type "NodePort" }}
-    nodePort: {{ .Values.server.observability.metrics.nodePort }}
-    {{- end }}
   {{- end }}
   selector:
     {{- include "kiali-server.selectorLabels" . | nindent 4 }}
--- a/kiali-server/templates/service.yaml
+++ b/kiali-server/templates/service.yaml
@@ -34,11 +34,17 @@ spec:
   {{- end }}
     protocol: TCP
     port: {{ .Values.server.port }}
+    {{- if eq .Values.deployment.service_type "nodePort" }}
+    nodePort: {{ .Values.server.nodePort }}
+    {{- end }}
   {{- if .Values.server.observability.metrics.enabled }}
   - name: http-metrics
     appProtocol: http
     protocol: TCP
     port: {{ .Values.server.observability.metrics.port }}
+    {{- if eq .Values.deployment.service_type "nodePort" }}
+    nodePort: {{ .Values.server.observability.metrics.nodePort }}
+    {{- end }}
   {{- end }}
   selector:
     {{- include "kiali-server.selectorLabels" . | nindent 4 }}
--- a/kiali-server/templates/service.yaml
+++ b/kiali-server/templates/service.yaml
@@ -35,7 +35,7 @@ spec:
     protocol: TCP
     port: {{ .Values.server.port }}
     {{- if eq .Values.deployment.service_type "NodePort" }}
-    nodePort: {{ .Values.server.nodePort }}
+    nodePort: {{ .Values.server.node_port }}
     {{- end }}
   {{- if .Values.server.observability.metrics.enabled }}
   - name: http-metrics
--- a/kiali-server/templates/service.yaml
+++ b/kiali-server/templates/service.yaml
@@ -34,7 +34,7 @@ spec:
   {{- end }}
     protocol: TCP
     port: {{ .Values.server.port }}
-    {{- if eq .Values.deployment.service_type "nodePort" }}
+    {{- if eq .Values.deployment.service_type "NodePort" }}
     nodePort: {{ .Values.server.nodePort }}
     {{- end }}
   {{- if .Values.server.observability.metrics.enabled }}
@@ -42,7 +42,7 @@ spec:
     appProtocol: http
     protocol: TCP
     port: {{ .Values.server.observability.metrics.port }}
-    {{- if eq .Values.deployment.service_type "nodePort" }}
+    {{- if eq .Values.deployment.service_type "NodePort" }}
     nodePort: {{ .Values.server.observability.metrics.nodePort }}
     {{- end }}
   {{- end }}
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -118,5 +118,4 @@ server:
     metrics:
       enabled: true
       port: 9090
-      nodePort: 32476
   web_root: ""
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -113,7 +113,7 @@ login_token:
 
 server:
   port: 20001
-  nodePort: 32475
+  node_port: 32475
   observability:
     metrics:
       enabled: true
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -113,6 +113,7 @@ login_token:
 
 server:
   port: 20001
+  #node_port:
   observability:
     metrics:
       enabled: true
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -113,7 +113,6 @@ login_token:
 
 server:
   port: 20001
-  node_port: 32475
   observability:
     metrics:
       enabled: true
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -113,8 +113,10 @@ login_token:
 
 server:
   port: 20001
+  nodePort: 32475
   observability:
     metrics:
       enabled: true
       port: 9090
+      nodePort: 32476
   web_root: ""


kiali-server/templates/service.yaml
```diff
@@ -34,11 +34,17 @@ spec:
   {{- end }}
     protocol: TCP
     port: {{ .Values.server.port }}
+    {{- if eq .Values.deployment.service_type "nodePort" }}
```
jmazzitelli, line 37: the value to look for should be `NodePort` (capital "N"). As per k8s docs [here](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport), let's maintain the same case as the official k8s docs (even if "nodePort" actually works; let's use what the docs say to use).

kiali-server/templates/service.yaml
```diff
@@ -34,11 +34,17 @@ spec:
...
+    nodePort: {{ .Values.server.nodePort }}
+    {{- end }}
   {{- if .Values.server.observability.metrics.enabled }}
   - name: http-metrics
     appProtocol: http
     protocol: TCP
     port: {{ .Values.server.observability.metrics.port }}
+    {{- if eq .Values.deployment.service_type "nodePort" }}
```
jmazzitelli, line 45: same as above - check for equality to `NodePort` (capital "N").

jmazzitelli (changes requested):

jmazzitelli:
Note to Kiali maintainers - do not merge this until an operator PR is submitted that provides the same feature. See: https://github.com/kiali/kiali/issues/7093#issuecomment-1927941164

This will require another community member or Kiali maintainer to take on that work.

kiali-server/templates/service.yaml
emreberber: Fixed. Thanks!

kiali-server/templates/service.yaml
emreberber: Fixed. Thanks!

kiali-server/templates/service.yaml
```diff
@@ -34,11 +34,17 @@ spec:
...
+    nodePort: {{ .Values.server.nodePort }}
+    {{- end }}
   {{- if .Values.server.observability.metrics.enabled }}
   - name: http-metrics
     appProtocol: http
     protocol: TCP
     port: {{ .Values.server.observability.metrics.port }}
+    {{- if eq .Values.deployment.service_type "NodePort" }}
+    nodePort: {{ .Values.server.observability.metrics.nodePort }}
+    {{- end }}
```
jmazzitelli, line 47: I don't think we want or need to expose the metrics over the nodePort. For one thing, it isn't used - the Prometheus scrape endpoint is still defined as using the service port as you see [here](https://github.com/kiali/helm-charts/blob/v1.80.0/kiali-server/templates/deployment.yaml#L31). Secondly, I don't see the need to expose the metrics endpoint over the nodePort - it is only to be used by Prometheus when scraping the metrics, and Prometheus should be able to do that over the normal port.

So I believe this should be removed and we should only expose the HTTP endpoint over the nodePort.

 ```suggestion
```

File: kiali-server/values.yaml [hunk 4]
```diff
@@ -113,8 +113,10 @@ login_token:
 
 server:
   port: 20001
+  nodePort: 32475
   observability:
     metrics:
       enabled: true
       port: 9090
+      nodePort: 32476
```
jmazzitelli, line 121: See my earlier comment about this value being unused and unnecessary.
```suggestion
```

jmazzitelli (changes requested):
I do not think the metrics endpoint should need to define a nodePort. See comments that follow.

kiali-server/values.yaml
```diff
@@ -113,8 +113,10 @@ login_token:
 
 server:
   port: 20001
+  nodePort: 32475
```
jmazzitelli, line 116: I also just noticed this is in camelCase. All settings within values and the Kiali CR are listed in snake_case. For consistency, this should be `node_port`:

```suggestion
  node_port: 32475
```

kiali-server/templates/service.yaml
```diff
@@ -34,11 +34,17 @@ spec:
   {{- end }}
     protocol: TCP
     port: {{ .Values.server.port }}
+    {{- if eq .Values.deployment.service_type "NodePort" }}
+    nodePort: {{ .Values.server.nodePort }}
```
jmazzitelli, line 38: use snake_case for consistency:
```suggestion
    nodePort: {{ .Values.server.node_port }}
```

jmazzitelli (changes requested):
use snake_case

jmazzitelli:
Another thing I just realized here - now that there is a fixed nodePort defined as a default, there is a greater chance of a collision and therefore deployment failure.

In current behavior (before this PR), k8s will pick an unused port and assign that as the service's nodePort. Now, with this PR, that is not the case. If there is something already on port 32475, the deployment will fail.

We need to maintain backwards compatibility - the behavior should be the same as before - a dynamic nodePort will be assigned by default. Only if the (optional) server.node_port is set will that be explicitly used for the node port.

That would retain backward compatibility and will reduce the chance for deployment failure in the default case, while at the same time still providing the ability for the user to configure the service with an explicit node port if that is desired.

jmazzitelli:
[Operator PR](https://github.com/kiali/kiali-operator/pull/739) is ready to go. Will wait for this PR to finish before the operator PR gets merged.

kiali-server/templates/service.yaml
```diff
@@ -34,6 +34,9 @@ spec:
   {{- end }}
     protocol: TCP
     port: {{ .Values.server.port }}
+    {{- if eq .Values.deployment.service_type "NodePort" }}
```
jmazzitelli, line 37: I haven't tested this but you will want something like this:

```suggestion
    {{- if and (not (empty .Values.server.node_port)) (eq .Values.deployment.service_type "NodePort") }}
```

kiali-server/values.yaml
```diff
@@ -113,6 +113,7 @@ login_token:
 
 server:
   port: 20001
+  node_port: 32475
```
jmazzitelli, line 116: Remove this, since we want the default to be undefined.
```suggestion
```

jmazzitelli (changes requested):
Waiting on one further update. See https://github.com/kiali/helm-charts/pull/248#issuecomment-1947217318 for the motivation behind this.

My suggestions to implement that are below.

kiali-server/values.yaml
emreberber: Do we still want to see the variable in values ? If it is okay for you, we can proceed in this way, of course

kiali-server/templates/service.yaml
emreberber: Do you think we should check if it is between **30000-32768** for the port number? I guess you did not check it on the operator side either. I wanted to ask 🤔

kiali-server/values.yaml
jmazzitelli: Actually, that's a good point. As the comments at the top of this values.yml file says:
```
# Settings that mimic the Kiali CR which are placed in the ConfigMap.
```

So values that the helm chart actually reference or use are documented here, even if the default value is undefined. So, yeah, can you just comment that out without a value. We do this in several other places in here, just mimic what is done like these:
* https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L63
* https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L98-L99

kiali-server/templates/service.yaml
jmazzitelli: We don't need to check the value of the node port (we'll let k8s fail if it is outside of its valid range) but we do want to check that node_port is not empty before setting it - this retains the backward compat. behavior.

emreberber: Yes, you are right, I thought so for a moment, but then it didn't make sense to me either, I deleted the message. 😄

jmazzitelli (approved):
Code looks good. Did a couple quick smoke tests:

1. Set node port to 32444:
```
helm install --namespace istio-system --set deployment.service_type=NodePort --set server.node_port=32444 kiali-server _output/charts/kiali-server-1.81.0-SNAPSHOT.tgz
```
results in:
```
$ kubectl get svc -n istio-system -l app.kubernetes.io/name=kiali -ojsonpath='{..spec.type}{"\n"}'
NodePort
```
```
$ kubectl get svc -n istio-system -l app.kubernetes.io/name=kiali -ojsonpath='{..spec.ports[?(@.name=="http")].nodePort}{"\n"}'
32444
```
2. Re-install, but this time do not specify node_port - old behavior remains - it chooses a free random port:
```
$ helm uninstall --namespace istio-system kiali-server
$ helm install --namespace istio-system --set deployment.service_type=NodePort kiali-server _output/charts/kiali-server-1.81.0-SNAPSHOT.tgz
$ kubectl get svc -n istio-system -l app.kubernetes.io/name=kiali -ojsonpath='{..spec.type}{"\n"}'
NodePort
$ kubectl get svc -n istio-system -l app.kubernetes.io/name=kiali -ojsonpath='{..spec.ports[?(@.name=="http")].nodePort}{"\n"}'
32515
```
3. Finally, don't set the service_type either:
```
$ helm uninstall --namespace istio-system kiali-server
$ helm install --namespace istio-system kiali-server _output/charts/kiali-server-1.81.0-SNAPSHOT.tgz
$ kubectl get svc -n istio-system -l app.kubernetes.io/name=kiali -ojsonpath='{..spec.type}{"\n"}'
ClusterIP
$ kubectl get svc -n istio-system -l app.kubernetes.io/name=kiali -ojsonpath='{..spec.ports[?(@.name=="http")]}{"\n"}' | jq
{
  "appProtocol": "http",
  "name": "http",
  "port": 20001,
  "protocol": "TCP",
  "targetPort": 20001
}
```

PR #250, opened 2024-03-04
feat(oidc): allow for secret csi driver mounts

rumstead:
https://github.com/kiali/kiali/issues/6942

To deploy the token in a scalable and secure manner, we want to leverage the secret CSI drivers backed by secret management tools. The secret CSI driver uses a pod's service account for authorization and will [only create the secret when there is a mount requested](https://secrets-store-csi-driver.sigs.k8s.io/topics/sync-as-kubernetes-secret). However, the kiali helm chat doesn't allow us to add any volume mounts which in turn doesn't trigger the secret creation.

The goal of this PR is to allow a user to add additional mounts and leverage the existing [custom_secrets](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.custom_secrets) configuration option.


kiali-server/values.yaml
```diff
@@ -87,7 +87,7 @@ deployment:
   tolerations: []
   version_label: ${HELM_IMAGE_TAG} # v1.39 # v1.39.0 # see: https://quay.io/repository/kiali/kiali?tab=tags
   view_only_mode: false
-
+  csiVolumes: []
```
rumstead, line 90: Example yaml with a defintion 
```yaml
deployment:
  csiVolumes:
    csi:
      driver: secrets-store.csi.k8s.io
      readOnly: true
      volumeAttributes:
        secretProviderClass: kiali-secretprovider
```

jmazzitelli: keep everything consistent with the way the rest of the current config is formatted - use snake_case for setting names.

here's our developer checklist for more things to keep in mind: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#how-to-change-resources-or-configuration

jmazzitelli: > keep everything consistent with the way the rest of the current config is formatted - use snake_case for setting names.

I say that, however, if I'm not mistaken, is everything underneat csiVolumes.csi all the yaml that gets copied directly in the volume definition of the Deployment? If so, they can probably stay since the "csi" is just defining the direct Deployment volume definition.

And I also think it looks like csiVolumes.csi is redundant. Why not just:

```
deployment.
  csi_volume:
    driver: ...
    readOnly: ...
    volumeAttributes: ...
```

And it is one volume, right? I don't see "csi" in your code as an array.. so `csi_volume` seems appropriate (as opposed to the plural `csi_volumes`.)

jmazzitelli: Oh, I think because you might need an "enabled" flag in here. Perhaps this is what we want?

```
csi:
  enabled: true | false
  csi_volume:
  ... your volume YAML here...
```

jmazzitelli: note: you don't HAVE to have an enabled flag - you could just say "if deployment.csi_volume" is not defined, assume there is no CSI and don't install any volume". That way you can just leave it with `deployment.csi_volume` - that is also backward compatible since no one has deployment.csi_volume  in their config today so it will jsut work as if it isn't defined/enabled.

rumstead: Thanks! Everything under the `csi` stanza maps to the csi [here](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/volume/#Volume). That should future-proof the helm chart against changes to the csi spec. If you prefer I template out every option, I could do that as well.

rumstead:
I still need to test this against a CSI Driver, I am just using other deployments to guide the yaml here. Once I test it, I will publish the PR and share some of the evidence.

kiali-server/values.yaml
rumstead: > And it is one volume, right? I don't see "csi" in your code as an array.. so `csi_volume` seems appropriate (as opposed to the plural `csi_volumes`.)

Correct it will most likely be a single volume. I made it an array to allow for additional mounts if required but maybe that is a premature optimization.

rumstead:
> I still need to test this against a secret csi driver, I am just using existing deployments that leverage the secret csi driver to guide the yaml here. Once I test it, I will publish the PR and share some of the evidence.

```
❯ vault kv get secret/app/kiali/oidc_secret
========== Secret Path ==========
secret/data/app/kiali/oidc_secret

======= Metadata =======
Key                Value
---                -----
created_time       2024-03-08T18:50:46.321771966Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            1

==== Data ====
Key      Value
---      -----
kiali    hello from vault kiali

❯ kubectl get secret kiali -o yaml
apiVersion: v1
data:
  oidc-secret: aGVsbG8gZnJvbSB2YXVsdCBraWFsaQ==
kind: Secret
metadata:
  creationTimestamp: "2024-03-08T18:50:50Z"
  labels:
    secrets-store.csi.k8s.io/managed: "true"
  name: kiali
  namespace: cnp-istio
  ownerReferences:
  - apiVersion: apps/v1
    kind: ReplicaSet
    name: kiali-59d9667b5b
    uid: 00ad1a4c-0a49-44ae-bc57-818d1435e459
  resourceVersion: "13121"
  uid: 49012f2c-cefd-4b22-bb29-de59d5befd95
type: Opaque

❯ k get deploy kiali -o yaml | yq '[.spec.template.spec.containers[0].volumeMounts, .spec.template.spec.volumes]'
## volumeMounts
- - mountPath: /kiali-configuration
    name: kiali-configuration
  - mountPath: /kiali-cert
    name: kiali-cert
  - mountPath: /kiali-secret
    name: kiali-secret
  - mountPath: /kiali-cabundle
    name: kiali-cabundle
  - mountPath: /mnt/secrets-store
    name: kiali-secret-csi
## volumes
- - configMap:
      defaultMode: 420
      name: kiali
    name: kiali-configuration
  - name: kiali-cert
    secret:
      defaultMode: 420
      optional: true
      secretName: istio.kiali-service-account
  - name: kiali-secret
    secret:
      defaultMode: 420
      optional: true
      secretName: kiali
  - configMap:
      defaultMode: 420
      name: kiali-cabundle
      optional: true
    name: kiali-cabundle
  - csi:
      driver: secrets-store.csi.k8s.io
      readOnly: true
      volumeAttributes:
        secretProviderClass: kiali-secretprovider
    name: kiali-secret-csi
```

kiali-server/values.yaml
```diff
@@ -87,7 +87,6 @@ deployment:
   tolerations: []
   version_label: ${HELM_IMAGE_TAG} # v1.39 # v1.39.0 # see: https://quay.io/repository/kiali/kiali?tab=tags
   view_only_mode: false
-
```
jmazzitelli, line 90: don't delete this line - notice in this values yaml we have a blank line to separate the major groupings of settings so it is easy to visually see.

jmazzitelli (changes requested):
wow, is this all we need :) I have a trivial change request. But if this is all you need, then I can do the operator changes.

jmazzitelli:
@rumstead Can you provide simple test procedures so I can manually test this? Doesn't have to mean installing CSI -- I just want to see what you used for the helm values that got the mounts to be defined in the deployment (even if it means without CSI that my deployment doesn't start). A dry-run of helm is fine, too.

I think you are using this for values, but I'm not sure:

```yaml
deployment:
  custom_secrets:
    csi:
      driver: secrets-store.csi.k8s.io
      readOnly: true
      volumeAttributes:
        secretProviderClass: "kiali-secretprovider"
```

rumstead:
> @rumstead Can you provide simple test procedures so I can manually test this? Doesn't have to mean installing CSI -- I just want to see what you used for the helm values that got the mounts to be defined in the deployment (even if it means without CSI that my deployment doesn't start). A dry-run of helm is fine, too.
> 
> I think you are using this for values, but I'm not sure:
> 
> ```yaml
> deployment:
>   custom_secrets:
>     csi:
>       driver: secrets-store.csi.k8s.io
>       readOnly: true
>       volumeAttributes:
>         secretProviderClass: "kiali-secretprovider"
> ```

Sure thing, here is the values.yaml I used called `local-values.yaml`. 

`helm template kiali-server -f local-values.yaml` 

```yaml
auth:
  strategy: "openid"
  openid:
    client_id: "abc-123"
    issuer_uri:  "https://foo"
    http_proxy: "http://proxy-foo"
    https_proxy: "http://proxy-foo"
    insecure_skip_verify_tls: true
    disable_rbac: true
deployment:
  image_version: v1.76.0
  version_label: v1.76.0
  custom_secrets:
    - name: "kiali-secret-csi"
      mount: "/mnt/secrets-store"
      csi:
        driver: secrets-store.csi.k8s.io
        readOnly: true
        volumeAttributes:
          secretProviderClass: kiali-secretprovider
server:
  web_port: 443
  web_schema: https
```

jmazzitelli:
Here's a quick test:

1. Make sure you pull this PR and build the helm charts via `make build-helm-charts`
2. Run this command:
```bash
helm template \
  --show-only templates/deployment.yaml \
  --set deployment.custom_secrets[0].name=csi-secret-name \
  --set deployment.custom_secrets[0].mount=/csi-secret-mount \
  --set deployment.custom_secrets[0].csi.driver=csi-secret-driver \
  --set deployment.custom_secrets[0].csi.readOnly=true \
  --set deployment.custom_secrets[0].csi.volumeAttributes.secretProviderClass=csi-secret-providerclass \
  kiali-server \
  _output/charts/kiali-server-1.82.0-SNAPSHOT.tgz
```

Essentially, the values that command passes in are these:
```yaml
deployment:
  custom_secrets:
  - name: "csi-secret-name"
    mount: "/csi-secret-mount"
    csi:
      driver: csi-secret-driver
      readOnly: true
      volumeAttributes:
        secretProviderClass: csi-secret-providerclass
```

3. Examine the Deployment yaml that results and see that it contains the CSI information.

volume mount:
```yaml
          volumeMounts:
...
            - name: csi-secret-name
              mountPath: "/csi-secret-mount"
```

volume:
```yaml
      volumes:
...
        - name: csi-secret-name
          csi:
            driver: csi-secret-driver
            readOnly: true
            volumeAttributes:
              secretProviderClass: csi-secret-providerclass
```

jmazzitelli:
@rumstead please confirm that test procedure I documented in the previous comment is correct and that it produces what you expect.

rumstead:
> @rumstead please confirm that test procedure I documented in the previous comment is correct and that it produces what you expect.

I get the expected output:

```yaml
        volumeMounts:
...
        - name: csi-secret-name
          mountPath: "/csi-secret-mount"


volumes:
...
      - name: csi-secret-name
        csi: 
          driver: csi-secret-driver
          readOnly: true
          volumeAttributes:
            secretProviderClass: csi-secret-providerclass
...

**Comment by jmazzitelli** [comment]
@rumstead I'm doing more testing and I noticed something that doesn't look broken, but I think I will need to document somewhere.

Does the CSI secret configuration support the "optional" field like the normal secret config? I don't think it does (I just tried it and I get validation errors when I try).

In my test procedure - in the helm template command - add `--set deployment.custom_secrets[0].optional=true` and notice the optional field isn't in the generated yaml (which is expected, because looking at this PR, it doesn't use that field in the if-stmt when csi is defined).

So run this:

```
helm template --show-only templates/deployment.yaml --set deployment.custom_secrets[0].name=csi-secret-name --set deployment.custom_secrets[0].mount=/csi-secret-mount --set deployment.custom_secrets[0].csi.driver=csi-secret-driver --set deployment.custom_secrets[0].csi.readOnly=true --set deployment.custom_secrets[0].csi.volumeAttributes.secretProviderClass=csi-secret-providerclass --set deployment.custom_secrets[0].optional=true kiali-server _output/charts/kiali-server-1.82.0-SNAPSHOT.tgz
```

and you don't see the optional field. You can see what happens if you set the optional field on a non-CSI secret by adding these params to that above command:

```
--set deployment.custom_secrets[1].name=second-name --set deployment.custom_secrets[1].optional=true --set deployment.custom_secrets[1].mount=/second-mount
```

That results in:

```yaml
        - name: second-name
          secret:
            secretName: second-name
            optional: true
```

It does not look like the `csi` section supports an `optional` field (like the `secret` section does).

If that is true, that I should probably change [the docs here](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.custom_secrets[*].optional) to say this `optional` field is ignored/not used when using `csi`.

jmazzitelli:
After some quick manual testing of my own, this PR is ready to merge as is the [operator PR](https://github.com/kiali/kiali-operator/pull/748#issuecomment-1994344218)

rumstead:
> @rumstead I'm doing more testing and I noticed something that doesn't look broken, but I think I will need to document somewhere.
> 
> In my test procedure - in the helm template command - add `--set deployment.custom_secrets[0].optional=true` and notice the optional field isn't in the generated yaml (which is expected, because looking at this PR, it doesn't use that field in the if-stmt when csi is defined).
> 
> So run this:
> 
> ```
> helm template --show-only templates/deployment.yaml --set deployment.custom_secrets[0].name=csi-secret-name --set deployment.custom_secrets[0].mount=/csi-secret-mount --set deployment.custom_secrets[0].csi.driver=csi-secret-driver --set deployment.custom_secrets[0].csi.readOnly=true --set deployment.custom_secrets[0].csi.volumeAttributes.secretProviderClass=csi-secret-providerclass --set deployment.custom_secrets[0].optional=true kiali-server _output/charts/kiali-server-1.82.0-SNAPSHOT.tgz
> ```
> 
> and you don't see the optional field. You can see what happens if you set the optional field on a non-CSI secret by adding these params to that above command:
> 
> ```
> --set deployment.custom_secrets[1].name=second-name --set deployment.custom_secrets[1].optional=true --set deployment.custom_secrets[1].mount=/second-mount
> ```
> 
> That results in:
> 
> ```yaml
>         - name: second-name
>           secret:
>             secretName: second-name
>             optional: true
> ```
> 
> Does the CSI secret configuration support the "optional" field like the normal secret config? I don't think it does. I just tried it and I get validation errors when I try - edit the deployment and add `optional: true` to the `csi` field and I see this: `# deployments.apps "kiali" was not valid: * patch: Invalid value: "map[spec:map[template:map[spec:map[]]]]": strict decoding error: unknown field "spec.template.spec.volumes[4].csi.optional"`.
> 
> It does not look like the `csi` section supports an `optional` field (like the `secret` section does).
> 
> If that is true, then I should probably change [the docs here](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.custom_secrets%5B*%5D.optional) to say this `optional` field is ignored/not used when using `csi`.
> 
> (update: I added a small blurb in those docs in this commit in the operator PR: [kiali/kiali-operator@b8848eb](https://github.com/kiali/kiali-operator/commit/b8848eb72d6ee57eb688f4c98a6ff5ee4312e8d6))

yea great catch and shout. I agree, I don't think the `csi` section supports it since it would be strange to say I want the driver to mount this but optionally. Thanks for adding the optional snippet.

jmazzitelli (approved):
this is ready to go - approve.

PR #251, opened 2024-03-20
remove verbose_mode that has been deprecated

jmazzitelli:
It has been deprecated/superceded since v1.26 (4 years ago)

part of: https://github.com/kiali/kiali/issues/7208

### DIFF
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -23,14 +23,10 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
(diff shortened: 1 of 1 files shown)


jmazzitelli:
how to test:

1. `make build-helm-charts`
2. `helm template kiali kiali-server --show-only templates/deployment.yaml --set deployment.logger.log_level=error`

Look at the output - find the LOG_LEVEL env var definition, it should say "error". This means everything worked, and nothing broke with the removal of the verbose_mode support.

```yaml
        - name: LOG_LEVEL
          value: "error"
```

nrfox (approved):

PR #259, opened 2024-05-06
comment out the ossmc image check - we will not smoke test it.

jmazzitelli:
part of https://github.com/kiali/kiali/issues/6865


jshaughn (approved):
Seems fine, not even sure why we need the commented code but not a problem.

jmazzitelli:
> not even sure why we need the commented code but not a problem

1. In case we ever want to resurrect it, we just uncomment
2. In case someone wanted to add that without knowing the history, there is the link to the issue explaining why it is commented out.

Doesn't hurt anything, as you say; I'm going to leave it as a breadcrumb for someone in the future.

PR #261, opened 2024-05-15
fix(kiali-server): don't create unused ClusterRole

capuche2412:
PR to solve https://github.com/kiali/kiali/issues/7357

### DIFF
--- a/kiali-server/templates/role-viewer.yaml
+++ b/kiali-server/templates/role-viewer.yaml
@@ -1,3 +1,4 @@
+{{- if or (.Values.deployment.view_only_mode) (ne .Values.auth.strategy "anonymous") -}}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -92,3 +93,4 @@ rules:
   verbs:
   - get
 ...
+{{- end -}}
--- a/kiali-server/templates/role.yaml
+++ b/kiali-server/templates/role.yaml
@@ -1,3 +1,4 @@
+{{- if not (or (.Values.deployment.view_only_mode) (ne .Values.auth.strategy "anonymous")) -}}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -59,7 +60,7 @@ rules:
   - security.istio.io
   - extensions.istio.io
   - telemetry.istio.io
-  - gateway.networking.k8s.io 
+  - gateway.networking.k8s.io
   resources: ["*"]
   verbs:
   - get
@@ -99,3 +100,4 @@ rules:
   verbs:
   - get
 ...
+{{- end -}}


kiali-server/templates/role-viewer.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if or (.Values.deployment.view_only_mode) (ne .Values.auth.strategy "anonymous") -}}
```
jmazzitelli, line 1: anonymous mode can still be used with view_only_mode=true (in fact, I think many people use it that way).

I don't think auth.strategy setting should affect which role is created.

Can you explain the thinking because checking for anonymous mode?

kiali-server/templates/role.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not (or (.Values.deployment.view_only_mode) (ne .Values.auth.strategy "anonymous")) -}}
```
jmazzitelli, line 1: see my previous comment about auth.strategy.

jmazzitelli (dismissed):

kiali-server/templates/role-viewer.yaml
capuche2412: I get your point. 
The goal is to have the same behavior with the [roleBinding](https://github.com/kiali/helm-charts/blob/cb4aebc46dc90ed0daa766c30777632bbd61fd27/kiali-server/templates/rolebinding.yaml#L5).
So if the roleBinding is set to viewer role, the viewer role is created (same for full rights role)

jmazzitelli: Ah, right. I remember this now.

* https://github.com/kiali/kiali/issues/5611.

jmazzitelli:
Here's a quick test to see things working. Run "helm template" with these combinations of settings - the first three should result in the `kiali-viewer` Role being created and used in the RoleBinding whose name is also `kiali-viewer`. The last should not have any `kiali-viewer` resource generated or used.

1. `deployment.view_only_mode=false` , `auth.strategy=token`
2. `deployment.view_only_mode=true` , `auth.strategy=token`
3. `deployment.view_only_mode=true` , `auth.strategy=anonymous`
4. `deployment.view_only_mode=false` , `auth.strategy=anonymous`

```
$ helm template --set deployment.view_only_mode=false --set auth.strategy=token _output/charts/kiali-server-*.tgz | grep "kiali-viewer"
  name: kiali-viewer
  name: kiali-viewer
  name: kiali-viewer
```

```
$ helm template --set deployment.view_only_mode=true --set auth.strategy=token _output/charts/kiali-server-*.tgz | grep "kiali-viewer"
  name: kiali-viewer
  name: kiali-viewer
  name: kiali-viewer
```

```
$ helm template --set deployment.view_only_mode=true --set auth.strategy=anonymous _output/charts/kiali-server-*.tgz | grep "kiali-viewer"
  name: kiali-viewer
  name: kiali-viewer
  name: kiali-viewer
```

```
$ helm template --set deployment.view_only_mode=false --set auth.strategy=anonymous _output/charts/kiali-server-*.tgz | grep "kiali-viewer"
<empty output>
```

That last one will show a different behavior from the current helm chart. Because both roles are created today, that last one will result in a single "name: kiali-vewer" line (that's the Role that is created but not used).

You can also examine the output of one of the first three commands to see that no Role is created with the name "kiali" that has the write permissions. A quick way to see this is to grep for "ClusterRole" and see that only one is created now with this PR, but today's helm charts create two roles:

This PR results in:
```
$ helm template --set deployment.view_only_mode=true --set auth.strategy=token _output/charts/kiali-server-*.tgz | yq | grep '^kind: ClusterRole'
kind: ClusterRole
kind: ClusterRoleBinding
```

but master build (today's helm chart) results in:

```
$ helm template --set deployment.view_only_mode=true --set auth.strategy=token _output/charts/kiali-server-*.tgz | yq | grep '^kind: ClusterRole'
kind: ClusterRole
kind: ClusterRole
kind: ClusterRoleBinding
```
```

jmazzitelli:
I did a quick check of the `hack/istio/multicluster/kiali-prepare-remote-cluster.sh` script which uses the server helm chart to prepare remote cluster secrets in a multi-cluster environment, and it should still work. The relevant code is [here](https://github.com/kiali/kiali/blob/v1.84/hack/istio/multicluster/kiali-prepare-remote-cluster.sh#L115), and this does not look like it requires the old behavior of getting both roles created - it only looks for role or role-viewer depending on the view_only_mode setting passed in by the user.

PR #263, opened 2024-05-21
[kiali-server] adding ability to configure extra environment variables

jdgeisler:
We noticed CPU throttling in Kiali due to GOMAXPROCS not being configured correctly. In order to fix this, we had to set the GOMAXPROCS environment variable in Kiali, however the helm chart currently does not have the ability to set additional environment variables.

This MR is adding the `extraEnvs` field to the kiali-server helm chart so that extra environment variables can be set in Kiali in a flexible way. An example of configuring this is below:

```
extraEnvs:
  - name: GOMAXPROCS
    value: "2"
  - name: test
    value: "true"
```

It then templates into the deployment like this:

```
env:
- name: ACTIVE_NAMESPACE
   valueFrom:
     fieldRef:
       fieldPath: metadata.namespace
- name: LOG_LEVEL
   value: "info"
- name: LOG_FORMAT
   value: "text"
- name: LOG_TIME_FIELD_FORMAT
   value: "2006-01-02T15:04:05Z07:00"
- name: LOG_SAMPLER_RATE
   value: "1"
- name: GOMAXPROCS
   value: "2"
- name: test
   value: "true"
```

By default, it is set to `extraEnvs: []` so that no extra environment variables are passed.

### DIFF
--- a/kiali-server/templates/deployment.yaml
+++ b/kiali-server/templates/deployment.yaml
@@ -123,6 +123,9 @@ spec:
           value: "{{ .Values.deployment.logger.time_field_format }}"
         - name: LOG_SAMPLER_RATE
           value: "{{ .Values.deployment.logger.sampler_rate }}"
+        {{- with .Values.deployment.extraEnvs }}
+        {{- . | toYaml | nindent 8 }}
+        {{- end }}
         volumeMounts:
         - name: {{ include "kiali-server.fullname" . }}-configuration
           mountPath: "/kiali-configuration"
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -72,6 +72,7 @@ deployment:
     log_level: "info"
     time_field_format: "2006-01-02T15:04:05Z07:00"
     sampler_rate: "1"
+  extraEnvs: []
   node_selector: {}
   pod_annotations: {}
   pod_labels: {}


jshaughn:
Hi @jdgeisler , thanks for the submission!  Would you mind [creating an issue](https://github.com/kiali/kiali/issues/new/choose) to go with it, that helps us keep track of things.

jmazzitelli:
Yes. What @jshaughn says. Here's the contribution readme where we document this: https://github.com/kiali/kiali/blob/master/CONTRIBUTING.md#making-a-change

The reason why we don't like people submitting PRs without an issue is because we lose the ability to dialogue with the community about the feature before you spend time working on a PR. For example, the server helm chart is merely a [convenience installer for Kiali](https://kiali.io/docs/installation/installation-guide/install-with-helm/#introduction) - the suggested install method is the [operator](https://kiali.io/docs/installation/installation-guide/install-with-helm/#install-with-operator). As such, any changes that go into the server helm charts must also go into the operator so we retain feature parity (the server helm chart should not have additional features that the operator does not).

Please see the developer checklist for the different touch points a feature like this is going to need to update before we merge it into any release: 
* [Are You Altering a Kiali Server Resource?](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-resource)
* [Are You Altering a Kiali Server Configuration Setting?](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting)

kiali-server/templates/deployment.yaml
```diff
@@ -123,6 +123,9 @@ spec:
           value: "{{ .Values.deployment.logger.time_field_format }}"
         - name: LOG_SAMPLER_RATE
           value: "{{ .Values.deployment.logger.sampler_rate }}"
+        {{- with .Values.deployment.extraEnvs }}
```
jmazzitelli, line 126: For legacy reasons, all settings are snake_case, not camelCase. To remain consistent across all Kiali settings in the Kiali CR and server helm charts, use snake_case for any new settings.

Also to remain consistent with other analogous settings (such as existing settings of [`additional_labels`](https://github.com/kiali/helm-charts/blob/v1.84.0/kiali-server/values.yaml#L61) and [`additional_service_yaml`](https://github.com/kiali/helm-charts/blob/v1.84.0/kiali-server/values.yaml#L44)), extra_envs should be renamed `additional_envs`.

kiali-server/values.yaml
```diff
@@ -72,6 +72,7 @@ deployment:
     log_level: "info"
     time_field_format: "2006-01-02T15:04:05Z07:00"
     sampler_rate: "1"
+  extraEnvs: []
```
jmazzitelli, line 75: For legacy reasons, all settings are snake_case, not camelCase. To remain consistent across all Kiali settings in the Kiali CR and server helm charts, use snake_case for any new settings.

Also to remain consistent with other analogous settings (such as existing settings of [`additional_labels`](https://github.com/kiali/helm-charts/blob/v1.84.0/kiali-server/values.yaml#L61) and [`additional_service_yaml`](https://github.com/kiali/helm-charts/blob/v1.84.0/kiali-server/values.yaml#L44)), extra_envs should be renamed `additional_envs`.

jmazzitelli (changes requested):

kiali-server/templates/deployment.yaml
jmazzitelli: Also, [as per the developer checklist mentioned earlier](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting), "Ensure each group of settings are alphabetically sorted in all the files where the new setting was added."

So this new `additional_envs` should be placed on the line before [`additional_service_yaml` found here](https://github.com/kiali/helm-charts/blob/v1.84.0/kiali-server/values.yaml#L44).

jdgeisler:
Hey @jshaughn and @jmazzitelli 

Thank you for the information. I have created the issue https://github.com/kiali/kiali/issues/7375

If this is okay, I can go ahead and work on the operator PR too.

jshaughn:
Closing as the issue was resolved in v1.83.

PR #267, opened 2024-06-21
Option to skip rbac resources creation in kiali-operator chart

rgaduput:
part of https://github.com/kiali/kiali/issues/7460

### DIFF
--- a/kiali-operator/templates/clusterrole.yaml
+++ b/kiali-operator/templates/clusterrole.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.create }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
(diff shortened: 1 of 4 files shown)


jmazzitelli (changes requested):
please see if this is what you want:

https://github.com/kiali/helm-charts/blob/master/kiali-operator/values.yaml#L41-L45

jmazzitelli:
See the discussion I started in the issue related to this PR. As it stands now, we do not want to merge this PR since it essentially cripples the operator (it won't have any Cluster Role, Binding, or Service Account upon installation if you set the new flag to disable rbac creation).

We can continue the discussion in the [issue](https://github.com/kiali/kiali/issues/7460) to further explain what your use case is and how you want things to work. As it is, this PR is not going to be useful for the vast majority of users of the operator.

jmazzitelli:
see:
* https://github.com/kiali/kiali/issues/8491

PR #270, opened 2024-07-01
[kiali-operator] Add podSecurityContext for deployment

brainstorm82:
Added podSecurityContext for kiali operator deployment

### DIFF
--- a/kiali-operator/templates/deployment.yaml
+++ b/kiali-operator/templates/deployment.yaml
@@ -42,6 +42,10 @@ spec:
       nodeSelector:
       {{- toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
+      {{- if .Values.podSecurityContext }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- end }}
       containers:
       - name: operator
         image: "{{ .Values.image.repo }}{{ if .Values.image.digest }}@{{ .Values.image.digest }}{{ end }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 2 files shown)


jmazzitelli (changes requested):
When asking for enhancements to Kiali, please follow the community contribution process as defined here:

Note: there is [already a security context you can set on the container](https://github.com/kiali/helm-charts/blob/master/kiali-operator/values.yaml#L24) which [contain some default constraints when not set](https://github.com/kiali/helm-charts/blob/master/kiali-operator/templates/deployment.yaml#L53-L64). As per the contribution doc, please create an issue with a description of the request, your use-case, and why the current feature set doesn't support it.

aravol:
Noting that this is causing problems in our deployment, since defaults occaisionally set `securityContext: []` which prevents Kiali from accessing logs

brainstorm82:
Can you provide more information on your configuration. So the change would need to not create securtiyContext at all if not set. Since no securityContext is not the same as securityContext: {}

jmazzitelli:
Can someone create a github issue on this so we can discuss there? This affects more than just the operator - it affects the helm charts, too. See my previous comment on this. These are the types of conversations we should have in the issue and then we can submit PRs once we determine what needs to get done.

jshaughn:
Issue created: https://github.com/kiali/kiali/issues/7987

jmazzitelli:
> Issue created: [kiali/kiali#7987](https://github.com/kiali/kiali/issues/7987)

DOH! I JUST created one myself! 

https://github.com/kiali/kiali/issues/7988

PR #271, opened 2024-07-04
Upgrade github actions to Node 20

ferhoyos:
Currently, github actions use v3 of actions/checkout, which internally runs on Node 16 (https://github.com/actions/checkout/blob/v3/action.yml#L90).

This PR upgrades github actions to v4, which internally runs Node on 20 (https://github.com/actions/checkout/blob/v4/action.yml#L102)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,20 +1,24 @@
 name: Helm charts
+
 on:
   push:
     branches:
-      - master
+    - master
   pull_request:
     types:
-      - opened
-      - reopened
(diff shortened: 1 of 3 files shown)


jshaughn:
@ferhoyos , just noticed this has been sitting for a while.  is there a reason we don't move it forward?

ferhoyos:
@jshaughn Yes, a PR approval of any of the Kiali maintainers :-)

jshaughn (approved):
Looks OK to me!

PR #279, opened 2024-09-03
Grant kiali role list permissions for mutating webhooks

nrfox:
This is necessary to read tags in the cluster.


jmazzitelli (dismissed):
this looks good, don't forget to add these to the OLM CSVs, also (the operator needs these when installed by OLM, hence the need for OLM CSVs to get updated). See my comment in the operator PR for details.

jmazzitelli (approved):
seams to luk ok - problmes fixed

jmazzitelli (approved):
works for me, in conjunction with the server PR

PR #281, opened 2024-09-09
fix version

leandroberetta:


jmazzitelli:
I thought we wanted to go to .3 ?

leandroberetta:
> I thought we wanted to go to .3 ?

For the case of patch releases, the action will take the current version and bump it, and that will be the release version.

jmazzitelli (approved):
ah.. right. this gets incremented.

PR #283, opened 2024-09-12
explicitly set spec.version in the CR

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/7741


jmazzitelli:
this goes with https://github.com/kiali/kiali-operator/pull/818

PR #284, opened 2024-09-25
support namespaces with just numerals in their names

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/7773

To test:

### server helm chart

1. `make build-helm-charts`
2. `kubectl create ns 12345`
3. `helm install --namespace 12345 kiali-server _output/charts/kiali-server*.tgz`

It should install fine, with no errors from helm. The server won't start because its trying to pull an image that doesn't exist, but that is irrelevant for this issue. We just want to make sure helm does not fail.

You can uninstall with:
```
helm uninstall --namespace 12345 kiali-server
```

### operator helm chart

1. `make build-helm-charts`
2. `kubectl create ns 12345`
3. `helm install --namespace 12345 --set cr.create=true kiali-operator _output/charts/kiali-operator*.tgz`

It should install fine, with no errors from helm. The operator won't start because its trying to pull an image that doesn't exist, but that is irrelevant for this issue. We just want to make sure helm does not fail.

You can uninstall with deleting the CR first, then the operator:
```
kubectl delete kiali -n 12345 --all && helm uninstall -n 12345 kiali-operator
```

### DIFF
--- a/kiali-operator/templates/clusterrolebinding.yaml
+++ b/kiali-operator/templates/clusterrolebinding.yaml
@@ -8,7 +8,7 @@ metadata:
 subjects:
 - kind: ServiceAccount
   name: {{ include "kiali-operator.fullname" . }}
-  namespace: {{ .Release.Namespace }}
+  namespace: "{{ .Release.Namespace }}"
 roleRef:
   kind: ClusterRole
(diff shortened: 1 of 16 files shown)


aljesusg (approved):
LGFM

PR #292, opened 2024-10-18
Add the possiblity to do a major release

leandroberetta:
This PR adds the ability to do major releases.

Schedule runs are disabled for now.

This PR also sets the 2.0 version for the next Kiali major release on next Monday.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,9 @@
 name: Release
 
 on:
-  schedule:
+  #schedule:
   # Every Monday at 08:00 (UTC) - try to make sure it happens after all Kiali-related images are released
-  - cron: '00 8 * * MON'
+  #- cron: '00 8 * * MON'
   workflow_dispatch:
     inputs:
       release_type:
@@ -11,6 +11,7 @@ on:
         required: true
         type: choice
         options:
+        - major
         - minor
         - patch
       release_branch:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -122,7 +126,10 @@ jobs:
             NEXT_VERSION=$(python bump.py $RELEASE_TYPE $RELEASE_VERSION)
         elif [[ $RELEASE_TYPE == "minor" ]]
         then 
-            NEXT_VERSION=$(python bump.py $RELEASE_TYPE $RELEASE_VERSION)          
+            NEXT_VERSION=$(python bump.py $RELEASE_TYPE $RELEASE_VERSION)
+        elif [[ $RELEASE_TYPE == "major" ]]      
+        then
+            NEXT_VERSION=$(python bump.py "minor" $RELEASE_VERSION)
```
jmazzitelli, line 132: I have no idea if this is right or wrong, but this just caught my eye. Since release type is "major" here, shouldn't the bump.py be sent in "major" as its argument value?
```suggestion
            NEXT_VERSION=$(python bump.py "major" $RELEASE_VERSION)
```

leandroberetta: For that case, which is calculate the next version, the bump shouldn't be to 3.0, but to 2.1 instead, for that reason I'm setting a minor bump there.

jmazzitelli (approved):

PR #294, opened 2024-10-29
Managed-by should only be set by `helm` or similar at install time

bleggett:
1. `app.kubernetes.io/managed-by: Helm` is automatically applied to resources _installed_ by Helm, and does not need to be (read: should not be) manually templated. The label is an install-time label that should be applied by the tool actually installing the resources into the cluster.

This matters because `helm template` should (and by default will) produce YAML without this label, as it is unknown at template-time what will "manage" the resource.

2. Also fix the `makefile` so it doesn't try to download the wrong Helm bin on linux/arm64.

_(@jmazzitelli added the fixes line below)_
fixes: https://github.com/istio/istio/issues/53698

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,8 @@ clean:
 	        i386)   arch="386" ;; \
 	        i686)   arch="386" ;; \
 	        x86_64) arch="amd64" ;; \
-	        arm|arm64)    dpkg --print-architecture | grep -q "arm64" && arch="arm64" || arch="arm" ;; \
+	        arm) arch="arm" ;; \
+	        arm64|aarch64) arch="arm64" ;; \
 	    esac ;\
 	    cd "${OUTDIR}/helm-install" ;\
 	    curl -L "https://get.helm.sh/helm-${HELM_VERSION}-$${os}-$${arch}.tar.gz" > "${OUTDIR}/helm-install/helm.tar.gz" ;\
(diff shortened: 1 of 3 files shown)


jmazzitelli:
This looks OK. Before we review/merge, can you create a github issue and link this PR to it so it can be tracked?

Please see the contributor's guide for details: https://github.com/kiali/kiali/blob/master/CONTRIBUTING.md#making-a-change

jmazzitelli (approved):
LGTM.

to test:

1. build the helm charts: `make build-helm-charts`
2. see that `helm template` output no longer adds the managed-by label:

```
helm template -n istio-system kiali-server _output/charts/kiali-server-*.tgz | grep managed-by
```
3. Install using `helm install` and see that it does include the managed-by label:
```
helm install --namespace istio-system kiali-server _output/charts/kiali-server-*.tgz
```

```
$ kubectl get -n istio-system clusterroles,deployment,cm,sa -l app.kubernetes.io/name=kiali -oyaml | grep Helm
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/managed-by: Helm
```

4. Do the same with the operator chart

```
helm template -n default kiali-operator _output/charts/kiali-operator-*.tgz | grep managed-by
```
```
helm install --namespace default kiali-operator _output/charts/kiali-operator-*.tgz
```
```
$ kubectl get -n default clusterroles,deployment,sa -l app.kubernetes.io/name=kiali-operator -oyaml | grep Helm
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/managed-by: Helm
      app.kubernetes.io/managed-by: Helm
```

PR #295, opened 2024-11-01
support remote_cluster_resources_only=true and support redirect_uris for openshift oauthclient

jmazzitelli:
this refactor the way we determine if it is an OpenShift cluster (for easier testing and maintenance)

part of https://github.com/kiali/kiali/issues/7861

To test quickly run this 4-part command:

```
echo -e "\nK8s full install:\n" && helm template -n istio-system --set deployment.ingress.enabled=true --set isOpenShift=false kiali-server _output/charts/kiali-server-*.tgz | grep "^kind:" && echo -e "\nOpenShift full install\n" && helm template -n istio-system --set deployment.ingress.enabled=true --set auth.openshift.redirect_uris[0]=foo --set isOpenShift=true kiali-server _output/charts/kiali-server-*.tgz | grep "^kind:" && echo -e "\nK8s remote only install:\n" && helm template -n istio-system --set deployment.ingress.enabled=true --set deployment.remote_cluster_resources_only=true --set isOpenShift=false kiali-server _output/charts/kiali-server-*.tgz | grep "^kind:" && echo -e "\nOpenShift remote only install\n" && helm template -n istio-system --set deployment.remote_cluster_resources_only=true --set deployment.ingress.enabled=true --set auth.openshift.redirect_uris[0]=foo --set isOpenShift=true kiali-server _output/charts/kiali-server-*.tgz | grep "^kind:"
```

and see that only some resources are created when remote_cluster_resources_only is true:

```

K8s full install:

kind: ServiceAccount
kind: ConfigMap
kind: ClusterRole
kind: ClusterRoleBinding
kind: Service
kind: Deployment
kind: Ingress

OpenShift full install

kind: ServiceAccount
kind: ConfigMap
kind: ConfigMap
kind: ClusterRole
kind: ClusterRoleBinding
kind: Service
kind: Deployment
kind: OAuthClient
kind: Route

K8s remote only install:

kind: ServiceAccount
kind: ConfigMap
kind: ClusterRole
kind: ClusterRoleBinding

OpenShift remote only install

kind: ServiceAccount
kind: ConfigMap
kind: ClusterRole
kind: ClusterRoleBinding
kind: OAuthClient
```

Notice things like the ingress/route, service, and deployment are not installed when we just want the remote cluster resources created.

### DIFF
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -20,11 +20,7 @@ Create chart name and version as used by the chart label.
 Determine if on OpenShift (when debugging the chart for OpenShift use-cases, set "simulateOpenShift")
 */}}
 {{- define "kiali-server.isOpenShift" -}}
-{{- if .Values.simulateOpenShift -}}
-true
-{{- else }}
-{{- .Capabilities.APIVersions.Has "operator.openshift.io/v1" -}}
-{{- end -}}
+{{- .Values.isOpenShift | default (.Capabilities.APIVersions.Has "operator.openshift.io/v1") -}}
 {{- end }}
 
 {{/*
--- a/kiali-server/templates/_helpers.tpl
+++ b/kiali-server/templates/_helpers.tpl
@@ -17,6 +17,17 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Determine if on OpenShift (when debugging the chart for OpenShift use-cases, set "simulateOpenShift")
+*/}}
+{{- define "kiali-server.isOpenShift" -}}
+{{- if .Values.simulateOpenShift -}}
+true
+{{- else }}
+{{- .Capabilities.APIVersions.Has "operator.openshift.io/v1" -}}
+{{- end -}}
+{{- end }}
+
+{{/*
 Identifies the log_level.
 */}}
 {{- define "kiali-server.logLevel" -}}
@@ -65,7 +76,7 @@ Determine the default web root.
     {{- .Values.server.web_root | trimSuffix "/" }}
   {{- end }}
 {{- else }}
-  {{- if .Capabilities.APIVersions.Has "route.openshift.io/v1" }}
+  {{- if eq "true" (include "kiali-server.isOpenShift" .) }}
     {{- "/" }}
   {{- else }}
     {{- "/kiali" }}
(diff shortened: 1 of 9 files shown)


jmazzitelli:
another test is to see the OAuthClient is not created in any case when auth.strategy is not openshift:

```
$ helm template -n istio-system --set auth.strategy=anonymous --set simulateOpenShift=x kiali-server _output/charts/kiali-server-*.tgz | grep "^kind:"
kind: ServiceAccount
kind: ConfigMap
kind: ConfigMap
kind: ClusterRole
kind: ClusterRoleBinding
kind: Service
kind: Deployment
kind: Route
```

```
$ helm template -n istio-system --set auth.strategy=anonymous --set deployment.remote_cluster_resources_only=true --set simulateOpenShift=x kiali-server _output/charts/kiali-server-*.tgz | grep "^kind:"
kind: ServiceAccount
kind: ConfigMap
kind: ClusterRole
kind: ClusterRoleBinding
```

kiali-server/templates/_helpers.tpl
```diff
@@ -16,6 +16,17 @@ Create chart name and version as used by the chart label.
...
+{{/*
+Determine if on OpenShift (when debugging the chart for OpenShift use-cases, set "simulateOpenShift")
+*/}}
+{{- define "kiali-server.isOpenShift" -}}
+{{- if .Values.simulateOpenShift -}}
+true
+{{- else }}
+{{- .Capabilities.APIVersions.Has "operator.openshift.io/v1" -}}
+{{- end -}}
+{{- end }}
```
nrfox, line 28: Instead of having another `simulateOpenShift` variable, we could just make `isOpenShift` (or maybe something like `platform: openshift`?) a variable and we only do the `.Capabilities.APIVersions.Has "operator.openshift.io/v1"` check if it's not defined?

jmazzitelli: See latest commit.

There really isn't any other option for "platform" .. it's either OpenShift or it's not. So I just made it look for "isOpenShift" and only do the .Capabilities check if its not defined.

I updated the test procedures in earlier comments to now use "isOpenShift=true" and "isOpenShift=false"... you can omit the `isOpenShift=false` since this will work without having to tell "helm template" that (because the .Capabilities will always return false anyway with "helm template").

nrfox (approved):

PR #296, opened 2024-11-06
Enable Custom tls config

PauloFerreira25:
Hello,
Please accept my collaboration.
I am creating the possibility of customizing the tls of the default kiali ingress.
Thank you for your attention

### DIFF
--- a/kiali-server/templates/ingress.yaml
+++ b/kiali-server/templates/ingress.yaml
@@ -37,6 +37,10 @@ spec:
   {{- if .Values.deployment.ingress.class_name }}
   ingressClassName: {{ .Values.deployment.ingress.class_name }}
   {{- end }}
+  {{- with .Values.deployment.ingress.tls }}
+  tls:
+    {{- tpl (toYaml .) $ | nindent 4 }}
+  {{- end }}
   rules:
   - http:
       paths:
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -54,6 +54,10 @@ deployment:
   image_pull_secrets: []
   image_version: ${HELM_IMAGE_TAG} # version like "v1.39" (see: https://quay.io/repository/kiali/kiali?tab=tags) or a digest hash
   ingress:
+    tls: [] # Custom tls config
+    #  - secretName: chart-example-tls
+    #    hosts:
+    #      - chart-example.local
     additional_labels: {}
     class_name: "nginx"
     #enabled:


jmazzitelli:
Here's the community contributors guidelines.. please follow these prior to submitting a PR:

(for example, we need an issue to discuss what enhancement is being asked for and what would need to change if the enhancement is to be accepted. Note that this PR just touches the server helm chart - but does nothing for the operator - and the operator must maintain feature parity with the server helm chart. We have a [checklist](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#developer-checklists-when-changing-resources-or-configuration) of things to help guide people, once it has been determined an enhancement request is warranted).

jmazzitelli:
_I would have put this in the github issue, had one been created prior to the PR being worked on_

This is just some custom setting for the ingress. The Kiali config does not provide all the knobs and switches to configure all the differrent Ingress possibilities, this is why override_yaml is provided. See:

https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.ingress.override_yaml

```
Because an Ingress into a cluster can vary wildly in its desired configuration, this setting provides a way to override complete portions of the Ingress resource configuration (Ingress on Kubernetes and Route on OpenShift). It is up to the user to ensure this override YAML configuration is valid and supports the cluster environment since the operator will blindly copy this custom configuration into the resource it creates.
```

Does the `override_yaml` not provide what you need to define your own Ingress definition with the tls settings?

PauloFerreira25:
Yes, it is possible to do this using overwrite, it would be possible to change everything.
However, additional optional fields are not an impact, but an improvement.
In addition to being used in other communities.

https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/ingress.yaml#L30C1-L33C13


The main impact of overwriting a flag is that improvements or fixes created by the official repository cannot break the overwrite, and therefore receive complaints.

But if my request does not benefit the project, you can close the issue.

jmazzitelli:
Right, in the case of Kiali, we originally did try to provide all the different settings that an ingress could possibly support, and it soon became apparent that we would never cover all of the different use cases of everyone's favorite ingress configs and even if we did, the Kiali ingress config would become large with an overbearing number of ingress flags (considering the Kiali CR is already large and overbearing :). So we instead introduced the override_yaml for this purpose.

So I am going to close this as it goes against the design philosophy of the Kiali ingress support, and it is already possible to do what is desired via that `override_yaml` config.

PR #299, opened 2024-11-13
allow customization of readiness probe settings

jshaughn:
kiali7832

### DIFF
--- a/kiali-server/templates/deployment.yaml
+++ b/kiali-server/templates/deployment.yaml
@@ -99,8 +99,8 @@ spec:
             {{- else }}
             scheme: HTTP
             {{- end }}
-          initialDelaySeconds: 5
-          periodSeconds: 30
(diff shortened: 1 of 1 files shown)


kiali-server/templates/deployment.yaml
```diff
@@ -99,8 +99,8 @@ spec:
             {{- else }}
             scheme: HTTP
             {{- end }}
-          initialDelaySeconds: 5
-          periodSeconds: 30
+          initialDelaySeconds: {{ .Values.server.readiness_probe.initial_delay_seconds | default 5 }}
+          periodSeconds: {{ .Values.server.readiness_probe.period_seconds | default 30 }}
```
jmazzitelli, line 103: these are more appropriate under .Values.deployment.
Define the defaults in the values.yaml file since they are being used here (that's why the PR test is red).

jmazzitelli (changes requested):

PR #303, opened 2024-12-16
increase signing key length

jmazzitelli:


jmazzitelli:
To test, build the helm charts and run "helm template" and see the generated (and random) signing_key now has a length of 32:

```
make build-helm-charts && helm template kiali-server _output/charts/kiali-server-*.tgz | grep signing_key
```

should return something like:

```
      signing_key: ipK0gNDt21nGujPwzivZZbKudF9dIGDY
```

Confirm that new random string is 32 characters long.

jshaughn (approved):
LG

PR #304, opened 2024-12-20
[operator] explicitly set metrics bind address to 8080

jmazzitelli:
the default value used to be 8080 but in ansible operator 1.36.0, that default was changed to 8443. So we need to explicitly set it 8080 now.

This goes with operator PR https://github.com/kiali/kiali-operator/pull/857


jmazzitelli:
Confirmed this works by the `metrics-test` molecule test that now passes when this helm chart PR is tested along with the operator PR https://github.com/kiali/kiali-operator/pull/857

nrfox (approved):

PR #95, opened 2021-08-05
Check Ingress is managed by networking.k8s.io/v1

nelson-lark:
Kubernetes version 1.18 `networking.k8s.io/v1` only manages `networkpolicies`. 

`Ingress` are still `networking.k8s.io/v1beta1` until 1.19+

This change ensures  Ingress exists in v1, else use v1beta1 logic.

```
kubectl version --short
Client Version: v1.21.3
Server Version: v1.18.20-eks-8c579e
```
```
 kubectl get --raw /apis/networking.k8s.io/v1 | jq .
{
  "kind": "APIResourceList",
  "apiVersion": "v1",
  "groupVersion": "networking.k8s.io/v1",
  "resources": [
    {
      "name": "networkpolicies",
      "singularName": "",
      "namespaced": true,
      "kind": "NetworkPolicy",
      "verbs": [
        "create",
        "delete",
        "deletecollection",
        "get",
        "list",
        "patch",
        "update",
        "watch"
      ],
      "shortNames": [
        "netpol"
      ],
      "storageVersionHash": "YpfwF18m1G8="
    }
  ]
}
```
Related to [kubernetes issue 90077](https://github.com/kubernetes/kubernetes/issues/90077)

PR #306, opened 2024-12-23
[kiali-server] Update ingress.yaml to fix apiVersion if condition

diwap:


jmazzitelli:
Please provide a github issue following the community contribution guidelines: 

Explain the bug you think you found and how you think it should be corrected. The Kiali maintainers can then have a dialogue with you before you spend time on code fixes that may or may not be needed.

Also note there is more than just the helm charts - bug fixes likely also need to be fixed in the operator (unless the operator also doesn't have the same bug - this is one reason we ask that community members create an issue so we can have a discussion about the issue prior to PRs being submitted).

jmazzitelli:
Note, in the github issue you will create, I will ask about this https://github.com/kiali/helm-charts/pull/95 (that `/Ingress` was ADDED in a previous release due to the reasons mentioned in that issue). So in your github issue, please address that PR #95 and why you think it was wrong to be added 3 years ago.

jmazzitelli (changes requested):
This partially reverts PR #95 (it changes one of the if-conditions but not the other one).

Do not merge until we have grounds for reverting PR #95.

jmazzitelli:
> Do we have an issue for this? If not we will soon close this PR.

There was no issue created, and the description of this PR is empty; thus we have no use-case or reason for partially reverting an older PR. I would close this.

jshaughn:
Closing, can re-open with a relevant issue.

PR #320, opened 2025-04-25
Added holdApplicationUntilProxyStarts in pod_annotations

hhovsepy:
Make sure pod_annotations `proxy.istio.io/config: '{ "holdApplicationUntilProxyStarts": true }'` is always set to be ready to install Kiali on auto-injection=enabled namespaces.

Issue https://github.com/kiali/kiali/issues/8259


hhovsepy:
Will these annotations be set when installing kiali-operator via helm-chart, and kiali-operator installs kiali? Not the case of installing kiali-server via helm-charts.

jmazzitelli (approved):

PR #323, opened 2025-06-05
kiali-server: disable RBACs if deployment.cluster_wide_access is false

karloramone:
# Problem

In clusters with dozens of Istio control plane namespaces, the Kiali Operator (using its Ansible playbooks) does not scale well and consumes an abnormal amount of memory.
To work around this issue, we have switched to using the kiali-server Helm chart instead.

However, it is currently not possible to disable the creation of ClusterRole and ClusterRoleBinding resources, which prevents us from managing RBAC externally — a requirement for our multi-tenancy setup.

# Proposed Change

This PR introduces a change that allows disabling RBAC creation when:
```
.Values.deployment.cluster_wide_access: false
```
This gives users more control over RBAC configuration and aligns better with environments that manage permissions externally.

### DIFF
--- a/kiali-server/templates/clusterrole-viewer.yaml
+++ b/kiali-server/templates/clusterrole-viewer.yaml
@@ -1,9 +1,10 @@
-{{- if not (or (.Values.deployment.view_only_mode) (ne .Values.auth.strategy "anonymous")) -}}
+{{- if .Values.deployment.cluster_wide_access -}}
+{{- if or (.Values.deployment.view_only_mode) (ne .Values.auth.strategy "anonymous") -}}
 ---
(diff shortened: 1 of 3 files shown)


jmazzitelli (changes requested):
~All this does is not create any roles at all. In order to support cluster_wide_access=false, the helm chart needs to create Roles in each of the "discovery selector" selected namespaces. Otherwise, Kiali is completely crippled with this PR as-is.~

I see what this is all about now. I'll add a comment below on this

jmazzitelli:
see:
* https://github.com/kiali/kiali/issues/8491

jmazzitelli:
implemented as part of https://github.com/kiali/helm-charts/pull/326

PR #338, opened 2025-08-14
verify operator permissions

jmazzitelli:


jmazzitelli:
when run after the [operator PR](https://github.com/kiali/kiali-operator/pull/949) is merged, this should work:
image

jmazzitelli:
The CI test fails currently because the operator PR has not been merged (the CI test will download the hack script from the operator repo so we don't have duplicate scripts). So once that operator PR is merged, the CI tests will be able to pass.

PR #339, opened 2025-08-20
add test framework and CI workflow

jmazzitelli:


jmazzitelli:
> Is the the same as https://helm.sh/docs/topics/chart_tests/?

no

PR #343, opened 2025-08-29
release workflow to validate crd sync

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/8681

this also checks the OSSMC CRD - we weren't doing that before


jmazzitelli:
I tested this in my fork and it seems to work. Now the helm charts CRDs are validated to be in sync with the kiali-operator's  golden copies.

PR #350, opened 2025-09-22
[CI] delay helm chart release workflow by one hour

jmazzitelli:
We want helm chart release to happen after all other release jobs are done and quay images are published.


jshaughn (approved):
LG

PR #393, opened 2026-04-07
Add x-kubernetes-preserve-unknown-fields to CRD

pbajjuri20:
Running the Molecule Pipeline, fixing the failure


jmazzitelli (changes requested):
don't change CRDs manually like this... using AI tool and load in AGENTS.md to help...

https://github.com/kiali/kiali/blob/master/AGENTS.md

In short, only change golden copy of CRDs (found in operator repo) then sync using make - requires operator and helm PRs

Aside from that, I don't know why this change is needed. This seems weird. Can you explain what this change is for? This should not be required.

pbajjuri20:
@jmazzitelli : "Just using this PR to run some quick Jenkins tests based on a change suggested by Cursor. No review needed—I'll be closing this in a few minutes!"

PR #394, opened 2026-04-16
Add kiali_feature_flags.authz.require_namespace_get config option

jshaughn:
Part-of: https://github.com/kiali/kiali/issues/7125

## Summary
- Syncs CRD with the new `authz.require_namespace_get` property from kiali/kiali-operator
- Adds `authz.require_namespace_get: false` default to `kiali-server/values.yaml`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kiali-server/values.yaml
+++ b/kiali-server/values.yaml
@@ -162,6 +162,8 @@ identity: {}
   #private_key_file:
 
 kiali_feature_flags:
+  authz:
+    require_namespace_get: false
   custom_workload_types: []
   disabled_features: []
   validations:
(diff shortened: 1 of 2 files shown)


kiali-server/values.yaml
```diff
@@ -162,6 +162,8 @@ identity: {}
   #private_key_file:
 
 kiali_feature_flags:
+  authz:
+    require_namespace_get: false
```
jmazzitelli: No need for this, revert these 2 additional lines.

For background, see the comment at the top of this values file:
```
```

This value is not used directly by the server helm chart, so no need to put it in here; "false" is already the default on server side.

```suggestion
```

jmazzitelli (changes requested):

jmazzitelli (approved):

PR #396, opened 2026-04-20
release/smoketest workflow retry

jmazzitelli:
The release workflows for kiali/kiali and kiali/kiali-operator push images to quay.io on independent schedules. This smoke test may run before those workflows finish, so we poll quay.io and wait for both images to appear before proceeding.

The retry occurs every 1 minute and times out in 90 minutes. This, along with the delayed schedule trigger for the smoke test workflow should help fix the release workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   schedule:
-  # Every Monday at 09:00 (UTC) - try to make sure it happens after all Kiali-related images are released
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -3,7 +3,7 @@ name: Release
 on:
   schedule:
   # Every Monday at 09:00 (UTC) - try to make sure it happens after all Kiali-related images are released
```
jshaughn:
```suggestion
  # Every Monday at 09:14 (UTC) - intentionally tries to be after all Kiali-related images are released, and avoids potential peak time of 9:00
```

jshaughn (approved):
LG

PR #408, opened 2026-05-27
feat: add OAuth2 credential secret scanning to Helm chart

emanuelbesliu:
## Summary

Adds OAuth2 `client_id` and `client_secret` credential secret scanning to the Helm chart's `process-auth-secrets` helper.

### `_helpers.tpl`
- Extended `process-auth-secrets` to extract OAuth2 `client_id` and `client_secret` when `.auth.oauth2` is present
- Uses the standard `secret:<name>:<key>` pattern with `value.txt` fileName

### Test
- Added `credential-secret-prometheus-oauth2.yaml` verifying volume + volumeMount generation

## Related PRs
- Kiali core: https://github.com/kiali/kiali/pull/9751
- Operator: https://github.com/kiali/kiali-operator/pull/1039

Assisted-by: Claude (Anthropic)


kiali-operator/crds/crds.yaml
```diff
@@ -1,4 +1,3 @@
----
```
jmazzitelli, line 1: The last and most recent commit removed this and the ending `...` seems to have been removed for no reason - I believe we want to keep these here for helm templating purposes. At any rate, there is no reason to remove these so I would say we want these back.

kiali-operator/crds/crds.yaml
```diff
@@ -2294,4 +2413,3 @@ spec:
                       See https://docs.openshift.com/container-platform/4.16/networking/routes/route-configuration.html#nw-configuring-route-timeouts_route-configuration for further details.
                     x-kubernetes-int-or-string: true
                     default: "60s"
-...
```
jmazzitelli, line 2325: restore `...` - no reason to remove it. i think it might actually be needed during some helm processing

jmazzitelli (changes requested):

jmazzitelli (changes requested):
Thanks for the Helm chart changes — I quickly compared against [the plan I posted in the discussion](https://github.com/kiali/kiali/discussions/9735#discussioncomment-17065061) and found several gaps.

## `client_id` should not use `secret:` notation

The `_helpers.tpl` `process-auth-secrets` helper runs `oauth2.client_id` through `extract-secret` alongside `oauth2.client_secret`. This is incorrect. `client_id` is a public identifier in the OAuth2 spec — it has no confidentiality requirement and does not belong in a Kubernetes Secret. No other Kiali auth config mounts `client_id` as a secret volume. The `client_id` extraction should be removed; only `client_secret` warrants this treatment.

## `client_secret` not validated at template time

The `validate-oauth2` helper checks for missing `client_id` and `token_url` but does not check `client_secret`. A user who sets `auth.type: oauth2` but omits `client_secret` gets no `helm template` error — the misconfiguration passes through silently. This applies to all five service validation blocks (prometheus, tracing, grafana, perses, custom_dashboards).

## Missing tests

The plan specified a full test matrix. The following are absent:

**Config rendering tests** — no tests verify that the oauth2 block is rendered correctly in the ConfigMap `config.yaml`:
- `auth-oauth2-prometheus-config.yaml`
- `auth-oauth2-grafana-config.yaml`
- `auth-oauth2-tracing-config.yaml`

**No-mount test** — no test verifies that a literal `client_secret` value (without the `secret:` prefix) does not create a volume:
- `auth-oauth2-literal-secret-no-volume.yaml`

**Disabled service test** — no test verifies that a disabled service with an oauth2 `secret:` reference does not create a volume:
- `credential-secret-oauth2-disabled-service.yaml`

**Backward compatibility tests** — no tests confirm that existing bearer and basic auth configurations are unaffected by these changes:
- `auth-oauth2-existing-bearer-unchanged.yaml`
- `auth-oauth2-existing-basic-unchanged.yaml`

**Individual misconfiguration failure tests** — the existing `oauth2-validation-missing-fields.yaml` bundles multiple scenarios together. The plan requires separate per-field tests so failures are unambiguous:
- `auth-oauth2-missing-client-secret.yaml`
- `auth-oauth2-missing-client-id.yaml`
- `auth-oauth2-missing-token-url.yaml`
- `auth-oauth2-missing-all-fields.yaml`

**Security guardrail test** — no test verifies that the oauth2 secret volume mount has `readOnly: true` enforced:
- `credential-secret-oauth2-security-guardrails.yaml`

jmazzitelli (changes requested):
Before posting your next update, please verify your own work against the plan using an AI tool with this prompt — this will tell you if you missed anything and is basically what I've been doing. It's best if you do this prior to requesting review:

> Read the implementation plan at https://github.com/kiali/kiali/discussions/9735#discussioncomment-17065061 in full. Then audit the code in this repo against every requirement in that plan. Report only what is wrong, missing, or incomplete — not what is correct. Be specific: file paths, line numbers, and what exactly is missing or broken. The full implementation spans three repos: kiali, kiali-operator, and helm-charts. Check all three.

Once the code fully implements the plan, I'll do a deeper review for correctness and security — but I can't do that until the plan is completely implemented.

---

I quickly compared the latest changes against [the plan I posted in the discussion](https://github.com/kiali/kiali/discussions/9735#discussioncomment-17065061). Several items are still incomplete.

## `use_kiali_token` conflict validation missing for perses and custom_dashboards

The `validate-oauth2` helper checks the `oauth2` + `use_kiali_token` incompatibility for prometheus, tracing, and grafana, but not for perses or custom_dashboards.prometheus. A user who sets both on either of those services gets no `helm template` error.

## Broken validation tests: missing `client_secret` in set params

Two existing `should_fail` tests will fail for the wrong reason:

- `oauth2-validation-tracing-grpc.yaml` — omits `client_secret` from the `--set` params, so the template fails on "client_secret is required" before it ever reaches the gRPC incompatibility check. The test passes by accident, not because the gRPC check works.
- `oauth2-validation-use-kiali-token.yaml` — same problem: omits `client_secret`, so the template fails on the required-fields check rather than the `use_kiali_token` check.

Both tests need `external_services.<service>.auth.oauth2.client_secret` added to their `--set` params so they actually exercise the validation path they claim to test.

## Missing config rendering tests

No tests verify that the oauth2 block is correctly rendered into the ConfigMap `config.yaml` for perses or custom_dashboards. The plan required one config rendering test per service; prometheus, grafana, and tracing have them, but perses and custom_dashboards do not.

kiali-server/templates/_helpers.tpl
```diff
@@ -509,6 +509,10 @@ Note: ca_file is deprecated in Kiali - use kiali-cabundle ConfigMap instead.
   {{- $result = merge $result (include "kiali-server.extract-secret" (dict "value" .auth.cert_file "volumeName" (printf "%s-cert" .prefix) "fileName" "useSecretKey") | fromJson) }}
   {{- /* Key file - uses secret key as filename */ -}}
   {{- $result = merge $result (include "kiali-server.extract-secret" (dict "value" .auth.key_file "volumeName" (printf "%s-key" .prefix) "fileName" "useSecretKey") | fromJson) }}
+  {{- /* OAuth2 client_secret (client_id is a public identifier and does not need secret mounting) */ -}}
+  {{- if .auth.oauth2 }}
```
jmazzitelli: **BUG-1: `process-auth-secrets` mounts oauth2 volume regardless of `auth.type`**

This block only checks `{{- if .auth.oauth2 }}` — it does not check `auth.type == "oauth2"`. If a user has `auth.type: basic` but also has an `oauth2` sub-block present (pre-staged config, leftover values), the secret volume will be mounted unnecessarily.

Fix: add `{{- if eq (toString .auth.type) "oauth2" }}` around the block.

**Severity: Important**

kiali-server/templates/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- include "kiali-server.validate-oauth2" . }}
```
jmazzitelli: **BUG-2: `validate-oauth2` fires before `remote_cluster_resources_only` guard**

`validate-oauth2` is invoked here unconditionally, before the `{{- if not .Values.deployment.remote_cluster_resources_only }}` guard on line 2. Remote-cluster-only deployments don't render the Deployment resource and don't use external services — yet the validation still runs and can reject valid shared values files. This is a commonly used deployment mode.

Fix: move the `{{- include "kiali-server.validate-oauth2" . }}` call to inside the `{{- if not .Values.deployment.remote_cluster_resources_only }}` guard.

**Severity: Important**

tests/kiali-server-tests/oauth2-validation-tracing-grpc.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: "Verify helm template fails when tracing uses oauth2 with use_grpc=true (default)"
+helm_args:
+  - "--set"
+  - "external_services.tracing.enabled=true"
+  - "--set"
+  - "external_services.tracing.auth.type=oauth2"
+  - "--set"
+  - "external_services.tracing.auth.oauth2.client_id=foo"
+  - "--set"
+  - "external_services.tracing.auth.oauth2.client_secret=secret::my-secret:key"
```
jmazzitelli: **BUG-3: Malformed `secret:` value with empty secretName**

`secret::my-secret:key` has a double colon — empty secretName between the first and second colon. The `extract-secret` helper matches `^secret:.+:.+` which requires at least one character between colons, so this value does not match and is treated as a literal string. The test still functions correctly for its intended purpose (validating the gRPC guard), but the value is malformed and misleading.

Fix: use `secret:my-secret:key` instead. Same issue in `oauth2-validation-use-kiali-token.yaml:9`.

tests/kiali-server-tests/oauth2-validation-use-kiali-token.yaml
```diff
@@ -0,0 +1,17 @@
+name: "oauth2_validation_use_kiali_token"
+description: "Verify helm template fails when oauth2 is combined with use_kiali_token"
+helm_args:
+  - "--set"
+  - "external_services.prometheus.auth.type=oauth2"
+  - "--set"
+  - "external_services.prometheus.auth.oauth2.client_id=foo"
+  - "--set"
+  - "external_services.prometheus.auth.oauth2.client_secret=secret::my-secret:key"
```
jmazzitelli: **BUG-3: Malformed `secret:` value with empty secretName**

`secret::my-secret:key` has a double colon — empty secretName. Same issue as noted in `oauth2-validation-tracing-grpc.yaml:11`. Fix: use `secret:my-secret:key`.

jmazzitelli (dismissed):

jmazzitelli (approved):

PR #410, opened 2026-05-28
[AI] Add tools include/exclude

josunect:
Include and exclude list of tools for Kiali chatbot. 

Ref. https://github.com/kiali/kiali/pull/9758

### DIFF
--- a/kiali-operator/crds/crds.yaml
+++ b/kiali-operator/crds/crds.yaml
@@ -233,6 +233,20 @@ spec:
                     description: "Enable or disable the ChatAI feature."
                     type: boolean
                     default: false               
+                  tools:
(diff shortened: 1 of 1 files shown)


aljesusg (dismissed):
LGFM

kiali-operator/crds/crds.yaml
```diff
@@ -233,6 +233,20 @@ spec:
                     description: "Enable or disable the ChatAI feature."
                     type: boolean
                     default: false               
+                  tools:
```
jmazzitelli: not alphabetical as per style guidelines - tools (T) should be after providers (P)

jmazzitelli (changes requested):

jmazzitelli (approved):

PR #415, opened 2026-06-03
Added new config option insecure_skip_verify for chat_ai

### LINKED ISSUE #9736: [AI] Make LightSpeed TLS verification configurable
The new LightSpeed provider currently creates its client with `WithInsecureSkipTLS(true)` unconditionally in `ai/providers/lightspeed/lightspeed_provider.go`.
That means TLS certificate verification is always disabled for LightSpeed HTTPS endpoints, even in normal production setups where the certificate should be validated.

### What is the new behavior?
TLS verification should be enabled by default.

If insecure TLS is needed, it should be controlled explicitly through configuration, for example:

a provider-specific insecure_skip_tls_verify option, or
reuse of an existing shared TLS/auth configuration pattern if there is one.

it would be more consistent to reuse Kiali's existing external-service auth/TLS model.

Kiali already supports outbound TLS settings for other integrations through config.Auth, including:

- insecure_skip_verify
- cert_file
- key_file

and builds the HTTP transport through the shared httputil.CreateTransport(...) helpers.

hhovsepy:
Core PR https://github.com/kiali/kiali/pull/9780
Fixes https://github.com/kiali/kiali/issues/9736

### DIFF
--- a/kiali-operator/crds/crds.yaml
+++ b/kiali-operator/crds/crds.yaml
@@ -254,6 +254,10 @@ spec:
                         endpoint:
                           description: "The endpoint of the provider. This is only used if a model has no endpoint specified."
                           type: string
+                        insecure_skip_verify:
+                          description: "Set true to skip verifying certificate validity when Kiali contacts the AI provider over https. When false (the default), the provider's TLS certificate is validated against the Kiali CA bundle."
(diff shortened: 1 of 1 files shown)


josunect (approved):
LGTM

PR #419, opened 2026-06-12
Add Gateway API HTTPRoute support to the kiali-server chart

### LINKED ISSUE #9876: Add Gateway API HTTPRoute support to the kiali install mechanisms (server helm chart/operator)
## Motivation

The Kubernetes SIG-Network and Security Response Committee have announced the [retirement of Ingress NGINX](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/) (best-effort maintenance until March 2026, then no further releases/bugfixes/CVE updates). The Gateway API is the SIG-Network-endorsed successor to Ingress — it is a standard upstream CRD set (`gateway.networking.k8s.io`), not an ingress-controller-specific configuration. As clusters migrate off Ingress NGINX, an increasing number of users will expose workloads through Gateway API `HTTPRoute` resources.

## Request

Add an **opt-in** `HTTPRoute` template to the `kiali-server` helm chart so the Kiali UI can be exposed through a Gateway, alongside the existing Ingress and OpenShift Route paths. Disabled by default; rendered only when explicitly enabled, attaching to a user-named Gateway.

A draft implementation already exists in kiali/helm-charts#419 (155/155 helm tests, `helm lint` clean) — opening this issue to formalize the request per CONTRIBUTING.

## Open design question for maintainers

@jmazzitelli noted on [discussions/9167](https://github.com/kiali/kiali/discussions/9167#discussioncomment-15675188) a preference against adding yet another separate ingress-style config block, and suggested extending the existing `spec.deployment.ingress` instead. Two points worth aligning on before any further work:

1. **Is the feature wanted at all?** Per CONTRIBUTING, work should proceed only after maintainers agree it belongs on the backlog — so this is the agreement checkpoint.
2. **If yes, what shape?** `HTTPRoute` is a distinct Gateway API CRD (a separate `kind`, not an Ingress), so it cannot be expressed cleanly through the Ingress `override_yaml` mechanism. That said, the chart-level UX could either (a) live under a small dedicated `httproute` block, or (b) be folded into the existing `deployment.ingress` section as a routing-backend selector. Happy to follow whichever direction the team prefers — or to close this out if the project would rather not take on a second routing surface.

Relates to: kiali/helm-charts#419

somaz94:
Fixes: kiali/kiali#9876

Adds an opt-in HTTPRoute (Gateway API) to the kiali-server chart so the Kiali UI can be exposed through a Gateway, alongside the existing Ingress and OpenShift Route paths. Disabled by default; when `deployment.httproute.enabled=true` it attaches to the Gateway(s) named in `deployment.httproute.parent_refs` and routes to the server Service on `server.port`, defaulting to a PathPrefix match on the Kiali web root (mirroring the Ingress behavior).

Validation:
- `make run-helm-tests TEST_SUITE=server` → 155/155 pass (153 existing + 2 new)
- New tests: `httproute-configuration` (parentRefs / hostnames / backend port / default path / labels) and `httproute-missing-parent-refs` (fail-fast guard when `parent_refs` is empty)
- `helm lint` clean

### DIFF
--- a/kiali-server/templates/httproute.yaml
+++ b/kiali-server/templates/httproute.yaml
@@ -1,0 +1,50 @@
+{{- if not .Values.deployment.remote_cluster_resources_only }}
+{{- if .Values.deployment.httproute.enabled }}
+{{- $route := .Values.deployment.httproute }}
+{{- if not $route.parent_refs }}
+{{- fail "deployment.httproute.enabled is true but deployment.httproute.parent_refs is empty; an HTTPRoute needs at least one parentRef to attach to a Gateway" }}
+{{- end }}
+---
+apiVersion: {{ $route.api_version | default "gateway.networking.k8s.io/v1" }}
+kind: HTTPRoute
+metadata:
+  name: {{ include "kiali-server.fullname" . }}
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    {{- if $route.additional_labels }}
+    {{- toYaml $route.additional_labels | nindent 4 }}
+    {{- end }}
+    {{- include "kiali-server.labels" . | nindent 4 }}
+  {{- with $route.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  parentRefs:
+    {{- toYaml $route.parent_refs | nindent 4 }}
+  {{- with $route.hostnames }}
+  hostnames:
+    {{- tpl (toYaml .) $ | nindent 4 }}
+  {{- end }}
+  rules:
+    {{- with $route.additional_rules }}
+    {{- tpl (toYaml .) $ | nindent 4 }}
+    {{- end }}
+  - backendRefs:
+    - name: {{ include "kiali-server.fullname" . }}
+      port: {{ .Values.server.port }}
+    matches:
+      {{- if $route.matches }}
+      {{- toYaml $route.matches | nindent 6 }}
+      {{- else }}
+    - path:
+        type: PathPrefix
+        value: {{ include "kiali-server.server.web_root" . }}
+      {{- end }}
+    {{- with $route.filters }}
+    filters:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
+...
+{{- end }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


jmazzitelli:
what issue is this addressing - need `fixes: <github URL>` referring to the issue in the description.

jmazzitelli (changes requested):
I disagree with this implementation. For a related enhancement request and my response to that (which is applicable to this), see https://github.com/kiali/kiali/discussions/9167#discussioncomment-15675188(comment)

somaz94:
Thanks for the review, @jmazzitelli — and fair point on process. I've opened kiali/kiali#9876 to formalize this and linked it via `Fixes:` above.

On the design concern: I want to be clear that Gateway API `HTTPRoute` isn't another ingress-controller-specific config block — it's the SIG-Network-endorsed successor to Ingress (a standard `gateway.networking.k8s.io` CRD), which is why I kept it strictly opt-in (`httproute.enabled=false` by default) and isolated from the existing Ingress path rather than expanding the Kiali CR's surface. Because it's a distinct `kind` (not an Ingress), it can't be expressed through the existing `ingress.override_yaml` mechanism.

That said, I'd rather follow your direction than push an unwanted surface. Two questions for the team on #9876:
1. Is HTTPRoute support something Kiali wants at all right now?
2. If so, would you prefer a small dedicated `httproute` block (as in this PR) or folding it into `deployment.ingress` as a routing-backend selector?

Happy to rework along whichever path you pick — or to close this out if the project would rather not take on a second routing surface. No hard feelings either way.

jmazzitelli:
I have added a comment to start the discussion over in the issue:
* https://github.com/kiali/kiali/issues/9876

somaz94:
Thanks for laying out the reasoning so clearly, that's a fair point and it matches the history you described.

The motivation on my end was just the Ingress NGINX retirement nudging users toward Gateway API, but I take your point that the route resource isn't really Kiali-specific, and that the override-section experience suggests this tends to grow into something users would rather hand-roll anyway. Since this is a project-direction call rather than an implementation detail, I'm happy to defer to you and the Kiali maintainers.

I'll close this for now. If the community discussion in #9876 lands on wanting it (with operator parity), I'm glad to pick it back up.

jshaughn:
Thanks, @somaz94 . As things progress and GW API adoption continues, maybe we can revisit. We'll see if other Kiali community members push for the same.

PR #420, opened 2026-06-16
Add option to add custom labels to the Service only

### LINKED ISSUE #9858: Add an option to add labels only to the Kiali service (i.e. via a new parameter service_labels)
The Kiali CR supports setting custom annotations on the Kiali Service via the `service_annotations` parameter. It would be useful to also have a parameter that allows you to add custom labels only to the Kiali Service and not any of the other resources. For example, a `service_labels` parameter.

Our use case is when we start using Solo's Istio distribution we need to label our services with custom solo labels for multi-cluster support.

### What is the current behavior?
It is currently possible to (ab)use the `.spec.deployment.extra_labels` parameter to do this, but it will label all resources and those labels do not make much sense there (although they do not cause any issues either).

### What is the new behavior?
With the extra parameter, custom labels can only be set on the Service resource.

rolandkool:
Fixes https://github.com/kiali/kiali/issues/9858

### DIFF
--- a/kiali-operator/crds/crds.yaml
+++ b/kiali-operator/crds/crds.yaml
@@ -877,6 +877,8 @@ spec:
                       ```
                       sidecar.istio.io/inject: "true"
                       ```
+                      Note that if you set a label key that conflicts with a standard Kiali label,
+                      your value will take precedence (user labels win on collision).
                     type: object
                     additionalProperties:
                       type: string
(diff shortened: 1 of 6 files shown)


jmazzitelli:
(I will be submitted a PR to this PR shortly to fix some issues - stay tuned)

jmazzitelli (changes requested):
This PR requires the changes that I submitted here:
* https://github.com/rolandkool/helm-charts/pull/1

jshaughn:
It seems this effort has been abandoned. Closing. Re-open as needed.

PR #426, opened 2026-07-10
Add opt-in OpenShift impersonation mode for multi-cluster auth

### LINKED ISSUE #10038: OpenShift Impersonation for Multi-Cluster Auth
## Problem

Kiali's multi-cluster support on OpenShift currently requires users to individually authenticate to each remote cluster via the Kiali UI's user profile dropdown. Each cluster login triggers a separate OpenShift OAuth redirect flow. This design is acceptable for small deployments (2-3 clusters) but becomes completely impractical at scale. An organization with 10, 50, or 100+ clusters cannot reasonably ask users to manually log into each one — the experience is so laborious that it effectively renders multi-cluster Kiali unusable for large fleets.

The root cause is that OpenShift OAuth does not provide SSO across clusters. Each cluster has its own OAuth server, its own tokens, and its own session. Kiali faithfully models this by maintaining per-cluster session cookies and requiring per-cluster authentication.

## Scope: OpenShift Only

Impersonation mode is scoped to the openshift auth strategy. Non-OpenShift environments already solve the multi-cluster single-login problem via the openid strategy with a shared OIDC provider — a single OIDC token is valid across all clusters, so there is no per-cluster login burden. Impersonation addresses the gap specific to OpenShift, where each cluster has its own independent OAuth server and there is no built-in cross-cluster SSO.

## Proposal

Add an opt-in impersonation mode for the OpenShift auth strategy. When enabled, the user authenticates once to the home cluster (where Kiali is deployed) via the normal OpenShift OAuth flow. For all cluster API calls (including the home cluster), Kiali then uses its own Service Account credentials (already provisioned on remote clusters via the existing remote cluster secret mechanism) and attaches Kubernetes impersonation headers (Impersonate-User, Impersonate-Group) carrying the authenticated user's identity. The remote cluster's API server authenticates the Kiali SA, verifies it has permission to impersonate, and then evaluates RBAC against the impersonated user — preserving per-user access control without requiring per-user tokens.

This reduces the user login experience from O(n) to O(1) regardless of cluster count, while maintaining the same security boundary: each user can only see and do what their RBAC on each remote cluster permits.

The feature is:

* Opt-in via a new Kiali CR configuration field (`spec.auth.openshift.impersonation.enabled`)
* OpenShift-only (non-OpenShift environments already solve this via the OpenID strategy with a shared OIDC provider)
* Backward-compatible (existing per-cluster OAuth flow remains the default)
* Security-conscious (impersonation privileges are only granted when explicitly opted in; the impersonated identity is derived server-side from the verified home cluster session, never from user-supplied input)

Note: For architectural consistency, impersonation is applied to all clusters including the home cluster. The user's OAuth token is used only for the one-time identity verification; all subsequent API calls use SA + impersonation.

PRs:
* server:  https://github.com/kiali/kiali/pull/10039
* operator: https://github.com/kiali/kiali-operator/pull/1078
* helm-charts: https://github.com/kiali/helm-charts/pull/426
* kiali.io docs: https://github.com/kiali/kiali.io/pull/992

jmazzitelli:
Fixes: https://github.com/kiali/kiali/issues/10038

When spec.auth.openshift.impersonation.enabled is true, users authenticate once to the home cluster and Kiali uses its SA credentials with Kubernetes impersonation headers for all cluster API calls. This eliminates the need for per-cluster OAuth login flows that don't scale beyond a few clusters.

Generated-by: Cursor


jshaughn (commented):
In line with the operator PR. Same basic comment from AI:

Same unconditional operator impersonate rule (consistent with operator PR). The server-side ClusterRole is properly conditional. 

One more thing I found:

The Helm chart contradicts the "OpenShift only" intent. The ClusterRole creation condition was changed to:

{{- $needsClusterRole := or (eq (include "kiali-server.auth.strategy" .) "openshift") (eq (include "kiali-server.deployment.tls_config.source" .) "auto") .Values.auth.openshift.impersonation.enabled }}

And there's a test (impersonation-clusterrole-present-triggers-creation.yaml) that explicitly validates that auth.strategy=anonymous + impersonation.enabled=true creates the OpenShift ClusterRole. That test is asserting the opposite of "OpenShift only."

jmazzitelli:
> Same unconditional operator impersonate rule (consistent with operator PR). The server-side ClusterRole is properly conditional.

As mentioned in the operator PR, this is to be expected. The operator must have permissions in order to grant them. Operator won't know if/when it needs to reconcile a Kiali CR with impersonation enabled so it needs the perms. So this is by design, no changes required. FWIW: this is mentioned in the docs PR - see "security considerations": https://github.com/kiali/kiali.io/pull/992/changes#diff-14cdaf68c76f32eb0646df31c6d5d281c075a42dce669df2625348cf1fa67df7R185

> The Helm chart contradicts the "OpenShift only" intent. The ClusterRole creation condition was changed to

True. That is a redundant if-clause expression that was added. Didn't break anything, but was not needed so it was removed and the test was removed since it validated settings that in reality would never have been.

jmazzitelli:
@jshaughn review comments address... ready for approval?

jshaughn (approved):
I don't see anything blocking this option, it seems valid and potentially useful for users that want to use it.

PR #429, opened 2026-07-29
Make the operator pod's probe timings configurable

### LINKED ISSUE #10132: Make the operator pod's probe timings configurable in the helm chart

Allow the Kiali operator pod's liveness/readiness/startup probe timings to be set from `values.yaml`.

### What is the current behavior?

The probe timings in `kiali-operator/templates/deployment.yaml` are hardcoded. The startup probe uses `initialDelaySeconds: 30`, so the operator is held unready for at least 30s even when it is actually up in a couple of seconds. On a local kind cluster that is a fixed 30s tax on every install.

### What is the new behavior?

A `probes` section in the operator chart values, with defaults identical to today's hardcoded values (no change for existing users):

```yaml
probes:
  liveness:
    periodSeconds: 30
  readiness:
    periodSeconds: 30
  startup:
    initialDelaySeconds: 30
    periodSeconds: 10
    failureThreshold: 6
```

Motivation: I redeploy a lab environment many times a day and would like to shorten that wait without forking the chart.

PR: kiali/helm-charts#429 (includes template tests covering both the defaults and overridden values).

h0tbird:
## Problem

The operator Deployment hardcodes its readiness, liveness and startup probe timings, so there is no way to tune them from the chart. `startupProbe.initialDelaySeconds: 30` in particular keeps the operator unready for at least 30 seconds no matter how quickly it actually starts, which dominates install time on fast clusters.

## Change

Moves the timings into a `probes.*` values block, mirroring what the Kiali CR already offers for the server via `spec.deployment.probes`:

```yaml
probes:
  liveness:
    periodSeconds: 30
  readiness:
    periodSeconds: 30
  startup:
    failureThreshold: 6
    initialDelaySeconds: 30
    periodSeconds: 10
```

The existing numbers become the defaults, so rendered output is unchanged unless a value is explicitly set.

Naming is camelCase to match the rest of `kiali-operator/values.yaml`. The server chart uses snake_case only because those values mirror the CR schema.

Closes https://github.com/kiali/kiali/issues/10132

## Testing

- `make clean build-helm-charts` — clean
- `./tests/run-helm-chart-tests.sh --test-suite operator` — 32/32 pass
- Two new tests: `deployment-probes-defaults.yaml` pins the current output so the defaults can't drift, `deployment-probes-custom.yaml` exercises the overrides

## Disclosure

Per the [AI policy](https://github.com/kiali/kiali/blob/master/AI_POLICY.md): this change was drafted with AI assistance and is being reviewed by me before it leaves draft.

### DIFF
--- a/kiali-operator/templates/deployment.yaml
+++ b/kiali-operator/templates/deployment.yaml
@@ -58,19 +58,19 @@ spec:
           httpGet:
             path: /readyz
             port: 6789
-          periodSeconds: 30
+          periodSeconds: {{ .Values.probes.readiness.periodSeconds | int | default 30 }}
         livenessProbe:
           httpGet:
             path: /healthz
             port: 6789
-          periodSeconds: 30
+          periodSeconds: {{ .Values.probes.liveness.periodSeconds | int | default 30 }}
         startupProbe:
           httpGet:
             path: /healthz
             port: 6789
-          initialDelaySeconds: 30
-          periodSeconds: 10
-          failureThreshold: 6
+          initialDelaySeconds: {{ .Values.probes.startup.initialDelaySeconds | int | default 30 }}
+          periodSeconds: {{ .Values.probes.startup.periodSeconds | int | default 10 }}
+          failureThreshold: {{ .Values.probes.startup.failureThreshold | int | default 6 }}
         securityContext:
         {{- if .Values.securityContext }}
         {{- toYaml .Values.securityContext | nindent 10 }}
--- a/kiali-operator/templates/deployment.yaml
+++ b/kiali-operator/templates/deployment.yaml
@@ -58,19 +58,19 @@ spec:
           httpGet:
             path: /readyz
             port: 6789
-          periodSeconds: {{ .Values.probes.readiness.periodSeconds | int | default 30 }}
+          periodSeconds: {{ .Values.probes.readiness.periodSeconds | int }}
         livenessProbe:
           httpGet:
             path: /healthz
             port: 6789
-          periodSeconds: {{ .Values.probes.liveness.periodSeconds | int | default 30 }}
+          periodSeconds: {{ .Values.probes.liveness.periodSeconds | int }}
         startupProbe:
           httpGet:
             path: /healthz
             port: 6789
-          initialDelaySeconds: {{ .Values.probes.startup.initialDelaySeconds | int | default 30 }}
-          periodSeconds: {{ .Values.probes.startup.periodSeconds | int | default 10 }}
-          failureThreshold: {{ .Values.probes.startup.failureThreshold | int | default 6 }}
+          initialDelaySeconds: {{ .Values.probes.startup.initialDelaySeconds | int }}
+          periodSeconds: {{ .Values.probes.startup.periodSeconds | int }}
+          failureThreshold: {{ .Values.probes.startup.failureThreshold | int }}
         securityContext:
         {{- if .Values.securityContext }}
         {{- toYaml .Values.securityContext | nindent 10 }}
--- a/docs/agents/kiali-operator-chart.md
+++ b/docs/agents/kiali-operator-chart.md
@@ -26,11 +26,6 @@ scribe:
       heading: Managed Kiali CR
   stale_flags: []
   review_notes:
-    - finding: "Startup probe parameters (initialDelaySeconds: 30, periodSeconds: 10, failureThreshold: 6) not documented — relevant for slow cluster deployments"
-      severity: minor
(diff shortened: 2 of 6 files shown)


jmazzitelli:
_(nm.. looks like you already know about the need to create an issue)_

h0tbird:
Yeah, I created https://github.com/kiali/kiali/issues/10132

h0tbird:
LGTM

kiali-operator/templates/deployment.yaml
```diff
@@ -58,19 +58,19 @@ spec:
...
         startupProbe:
           httpGet:
             path: /healthz
             port: 6789
-          initialDelaySeconds: 30
-          periodSeconds: 10
-          failureThreshold: 6
+          initialDelaySeconds: {{ .Values.probes.startup.initialDelaySeconds | int | default 30 }}
+          periodSeconds: {{ .Values.probes.startup.periodSeconds | int | default 10 }}
+          failureThreshold: {{ .Values.probes.startup.failureThreshold | int | default 6 }}
```
jmazzitelli: One issue with the `probes.*` implementation: the template uses `.Values.X | int | default N`, and `default` treats a converted `0` as "empty," so it silently falls back to the hardcoded default instead of honoring an explicit `0`. This only matters for `probes.startup.initialDelaySeconds` — `periodSeconds` and `failureThreshold` require a minimum of `1` per the Kubernetes API, so `0` isn't a legal value for those anyway and getting coerced back to the default is harmless. But `initialDelaySeconds: 0` is a legitimate and fairly standard tuning choice for a `startupProbe` specifically (it doesn't gate liveness/readiness, so there's no restart risk in probing immediately — you just let `periodSeconds × failureThreshold` be the real "how long to wait" budget), and it's exactly the value someone chasing this PR's own "fast clusters" motivation would reach for. Right now setting it silently does nothing, with no error or warning.

For the fix, since `values.yaml` already declares a default of `30` for this key, the `| default 30` in the template is redundant and is the actual source of the bug — just drop it. That relies on the chart's own `values.yaml` default (already `30`) rather than a second template-level fallback, so backward-compat behavior is unchanged, but an explicit `0` now renders as `0`.

In fact, I would drop all the defaults and just rely on the defaults set in values.yaml:

```suggestion
          initialDelaySeconds: {{ .Values.probes.startup.initialDelaySeconds | int }}
          periodSeconds: {{ .Values.probes.startup.periodSeconds | int }}
          failureThreshold: {{ .Values.probes.startup.failureThreshold | int }}
```

Also, add a new test case that sets `probes.startup.initialDelaySeconds=0` and asserts it renders as `0` (not `30`)

kiali-operator/templates/deployment.yaml
```diff
@@ -58,19 +58,19 @@ spec:
           httpGet:
             path: /readyz
             port: 6789
-          periodSeconds: 30
+          periodSeconds: {{ .Values.probes.readiness.periodSeconds | int | default 30 }}
```
jmazzitelli: Similar to earlier comment, remove `default` and let the values.yaml provide the defaults.

```suggestion
          periodSeconds: {{ .Values.probes.readiness.periodSeconds | int }}
```

kiali-operator/templates/deployment.yaml
```diff
@@ -58,19 +58,19 @@ spec:
...
         livenessProbe:
           httpGet:
             path: /healthz
             port: 6789
-          periodSeconds: 30
+          periodSeconds: {{ .Values.probes.liveness.periodSeconds | int | default 30 }}
```
jmazzitelli: Similar to earlier comment, remove `default` and let the values.yaml provide the defaults.

```suggestion
          periodSeconds: {{ .Values.probes.liveness.periodSeconds | int }}
```

jmazzitelli (dismissed):

jmazzitelli (commented):
These changes look good. I'll do a final manual review and approve/merge if all remains looking good.

kiali-operator/templates/deployment.yaml
h0tbird: Thanks! That makes sense, I’ve fixed it.

kiali-operator/templates/deployment.yaml
h0tbird: Tahnks! Fixed.

kiali-operator/templates/deployment.yaml
h0tbird: Fixed.

jmazzitelli (approved):
LGTM

PR #436, opened 2026-09-09
server helm chart - clean up the values.yaml

jmazzitelli:
some settings here are deprecated. others are not used by the templates and shouldn't be here


jshaughn (approved):
some root_namespace kruft, and the ai.chat prefix looks important.

PR #439, opened 2026-09-21
Sync Kiali CRD show_selector deprecation

ferhoyos:
## Summary

- Sync `kiali-operator/crds/crds.yaml` with kiali-operator CRD changes for deprecated `show_selector` (default `true`)

Companion to kiali/kiali-operator#1098 and kiali/kiali#10329.

## Test plan

- [ ] `make validate-crd-sync` in kiali-operator repo passes against this branch

Made with [Cursor](https://cursor.com)


kiali-operator/crds/crds.yaml
```diff
@@ -2284,9 +2284,9 @@ spec:
                             description: "Default language used in Kiali application."
                             type: string
                           show_selector:
-                            description: "If true Kiali masthead displays language selector icon."
+                            description: "Deprecated. Language selection is available in Preferences by default. If false, the language selector is hidden until this setting is removed in a future release."
```
jmazzitelli: follow the deprecation wording in the other settings for consistency. We use all-caps, we say which productized version is the last version it wasn't deprecated, followed by `:`. For example, we have this in places:

```yaml
description: "DEPRECATED AFTER v1.73: The URL of Kiali in the cluster."
```

You will probably want this to be:

```suggestion
                            description: "DEPRECATED AFTER v2.27: Language selection is available in Preferences by default. If false, the language selector is hidden until this setting is removed in a future release."
```

Also, I am not sure if changing the default value is allowed without causing up to add a new CRD schema version - confirm that (i.e. install a previous version where the CRD default was false, then create a Kiali CR without specifying show_selector, then install a new Kiali CRD that has the new default of true - does the CRD update fail or not?)

jmazzitelli (approved):

